PR #7, opened 2022-08-23
Added copyright notice in README and added missing ones in NOTICE.md file

shamsasari:


shamsasari:
As advised by the external counsel.

bonifazi-r3 (approved):

PR #8, opened 2022-08-23
[NOTICK] Fix broken links to the api docs on the docsite

ac101m:
# Changes
- Fix broken links to the api docs on the docsite


shamsasari (changes requested):
I think there's been a misunderstanding. There's no need to update the doc links. Just revert the config which controls the links to be `conclave` again. What I actually want is `Conclave Core` to appear in the title, like it does here: https://r3-conclave.artifacts.teamcity.com/presignedTokenAuth/d7b1bd54-87e9-43f7-b474-7da8aa407dd8/repository/download/ConclaveCore_SdkBuild/3649:id/docs.zip!/api/index.html

ac101m:
> I think there's been a misunderstanding. There's no need to update the doc links. Just revert the config which controls the links to be `conclave` again. What I actually want is `Conclave Core` to appear in the title, like it does here: https://r3-conclave.artifacts.teamcity.com/presignedTokenAuth/d7b1bd54-87e9-43f7-b474-7da8aa407dd8/repository/download/ConclaveCore_SdkBuild/3649:id/docs.zip!/api/index.html

I tried that, it doesn't work. The module name controls both the links and the title. Displayname doesn't seem to do anything that I can see :man_shrugging:

shamsasari:
The escaped space character is really ugly. No way to fix that is there? Perhaps rename to `Conclave-Core` to fix that. WDYT?

ac101m:
> The escaped space character is really ugly. No way to fix that is there? Perhaps rename to `Conclave-Core` to fix that. WDYT?

The links work fine without the escaped space, but unfortunately autocomplete of the URL in intellij doesn't

shamsasari:
Let's just get rid of the space. I know it's supported in browsers, but I just can't help thinking this will come back to bite us!

shamsasari (approved):

PR #9, opened 2022-08-24
CON-1020: Publish Conclave SDK to Maven Central

bonifazi-r3:
This patch contains the configuration fixes to publish Conclave SDK artifacts to Maven Central

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -349,8 +349,8 @@ for (publishedProject in publishedProjects) {
     project.signing {
         // Skip signing if the key is not available, which is what we want for local builds.
         //   Signing is required for releases, i.e. when we want to publish to Maven Central
-        // required = false means that this task will be executed only when username and passwords are available
-        //   i.e. this task does not run when building locally.
+        // "required = false" means that this task will only be executed when username and passwords are available
+        //   as environment variables, i.e. this task should not run when building localy.
         required = false
         def key = System.getenv("CONCLAVE_PUBLICATION_SIGNING_PRIVATE_KEY")
         def password = System.getenv("CONCLAVE_PUBLICATION_SIGNING_PRIVATE_KEY_PASSWORD")
--- a/build.gradle
+++ b/build.gradle
@@ -45,6 +45,10 @@ enum VersionType {
     SNAPSHOT
 }
 
+def gitBranch = getGitBranch()
+def ossRHUsername = System.getenv("CONCLAVE_OSSRH_USERNAME")
+def ossRHPassword = System.getenv("CONCLAVE_OSSRH_PASSWORD")
+
 ext {
     conclave_revision = getConclaveRevision()
     project.logger.info("Conclave revision: $conclave_revision")
@@ -53,8 +57,16 @@ ext {
         releaseType = VersionType.SNAPSHOT
     } else if (version.matches(/.+\-RC[0-9]+$/)) {
         releaseType = VersionType.RELEASE_CANDIDATE
-    } else {
+    } else if ((gitBranch == "release/$conclave_version") && ossRHUsername != null) {
+        // We consider this a GA release only if
+        //  - it is not a RELEASE_CANDIDATE or SNAPSHOT
+        //  - the current branch is a release branch,
+        //  - the related env variables are set (i.e. only from TeamCity)
         releaseType = VersionType.GA_RELEASE
+    } else {
+        //  This is to avoid that we publish by mistake to Maven Central
+
+        releaseType = VersionType.RELEASE_CANDIDATE
     }
 }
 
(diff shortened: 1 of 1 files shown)


bonifazi-r3:
@shamsasari Please have a first look and give feedback. I guess now it will be trial and error, not sure if we want to test this at the last stage (i.e. when we trigger the TC release/1.3 build). We can discuss this tomorrow.

build.gradle
```diff
@@ -343,17 +381,34 @@ for (publishedProject in publishedProjects) {
...
+                        developer {
+                            name = 'Conclave Team'
+                            email = "conclave@r3.com"
+                            organization = 'R3 LLC'
+                            organizationUrl = "https://www.r3.com"
+                        }
+                    }
+                    scm {
+                        connection = "scm:git:git://github.com/R3Conclave/conclave-core-sdk.git"
+                        developerConnection = "scm:git:ssh://github.com:R3Conclave/conclave-core-sdk.git"
```
shamsasari, line 410: According to the POM docs, this is only for write access, whatever that means. I think we should remove this one.

build.gradle
```diff
@@ -343,17 +381,34 @@ for (publishedProject in publishedProjects) {
             maven(MavenPublication) {
                 from project.components.java
                 pom {
+                    groupId = project.group
+                    artifactId = project.name
+                    version = project.version
+                    packaging = "jar"
```
shamsasari, line 387: These are all defaults and so there should be no need to specify them.

build.gradle
```diff
@@ -39,9 +39,34 @@ String getConclaveRevision() {
...
+
 ext {
     conclave_revision = getConclaveRevision()
     project.logger.info("Conclave revision: $conclave_revision")
+
+    if (version.endsWith("-SNAPSHOT")) {
+        artifactType = ArtifactType.SNAPSHOT
+    } else if (version.matches(/.+\-RC[0-9]+$/)) {
+        artifactType = ArtifactType.RELEASE_CANDIDATE
+    } else if ((gitBranch == "release/$conclave_version") && ossRHUsername != null) {
```
shamsasari, line 60: The version string should be enough to determine what the version type is, no need to double-check the branch.

build.gradle
```diff
@@ -39,9 +39,34 @@ String getConclaveRevision() {
     return conclave_revision
 }
 
+enum ArtifactType {
```
shamsasari, line 42: I think VersionType is better name. ArtifactType implies we're dealing with different artifacts, whatever that might mean (jar, war, etc).

shamsasari (changes requested):

build.gradle
bonifazi-r3: Ok, I struggled a bit to understand that and thought it was nice to be explicit. Will remove

build.gradle
bonifazi-r3: This is something I had issues with. I.e. when I was playing locally I had a current `release/1.3` branch (by mistake) and I had issues as the build was trying to run the sign task.
I also thought about the possibility where we branch out from the current `release/1.3` and push the changes. In that case, again, we would have issues as the build would try to sign and publish.
The checks above are narrowing the possibility of mistakes like mine

build.gradle
bonifazi-r3: Replaced

build.gradle
bonifazi-r3: [developers](https://central.sonatype.org/publish/requirements/#scm-information)
I understand they are compulsory, but we can put whatever we like there. But if the users do not have access, it does not seem to be a problem

build.gradle
shamsasari: We shouldn't be branching off release/1.3. Why would we do that? Patch releases (e.g. 1.3.1) will continue to be on release/1.3.

I'm not comfortable with this check. It changes the defintion of the version based on the branch and the local setup (i.e. ossRHUsername).

build.gradle
```diff
@@ -343,17 +381,29 @@ for (publishedProject in publishedProjects) {
             maven(MavenPublication) {
                 from project.components.java
                 pom {
+                    name = project.name
```
shamsasari, line 384: Also remove

build.gradle
```diff
@@ -271,6 +285,8 @@ task buildDocs(dependsOn: ['mkdocs'])
 dokkaHtml.shouldRunAfter mkdocs
 // endregion
 
+def projectURL = "https://github.com/R3Conclave/conclave-core-sdk"
```
shamsasari, line 288: We can inline this

shamsasari (commented):
Just one minor change, otherwise it's looking good.

shamsasari (approved):

PR #12, opened 2022-08-25
[CON-1025] Remove the handler framework used in the internal enclave-host comms

ac101m:
# Changes
- Delete unused channel handlers and associated tests.
- Implement the "call" framework as a replacement for the existing handler mechanism.
- Add `initialise` method to `EnclaveHandle` and remove lazy initialisation of enclaves (now initialised explicitly in `EnclaveHost.start`).


conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeHostCallInterface.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * Handle call initiations from the host.
+     * This method propagates the call to the appropriate enclave side call handler, then sanitises, serialises and
+     * propagates any exceptions that occur. If a return value is produced, a reply message is sent back to the host.
+     */
+    private fun handleCallEcall(callType: EnclaveCallType, parameterBuffer: ByteBuffer) {
+        try {
+            acceptCall(callType, parameterBuffer)?.let {
```
bonifazi-r3, line 91: Is there any possibility that the buffer coming out of this call is empty? In that case we  seem to do nothing

bonifazi-r3 (changes requested):
Looks very good, just a first round of comments

conclave-common/src/test/kotlin/com/r3/conclave/common/internal/CallHandlerTest.kt
```diff
@@ -0,0 +1,3 @@
+package com.r3.conclave.common.internal
+
+class CallHandlerTest {}
```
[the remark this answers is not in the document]
ac101m: Nope, accidentally committed, will remove

conclave-common/src/main/kotlin/com/r3/conclave/common/MockCallInterfaceConnector.kt
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * This class serves as the glue between a mock host and a mock enclave.
+ * An instance of this class exists "between" the mock enclave and the host, bridging the
+ * [com.r3.conclave.enclave.internal.HostCallInterface] to the [com.r3.conclave.host.internal.EnclaveCallInterface].
+ * Parameter and return value buffers are deep copied to more accurately emulate the behaviour of an actual enclave.
+ */
+class MockCallInterfaceConnector {
+    companion object {
+        private fun copyBuffer(buffer: ByteBuffer): ByteBuffer {
+            return ByteBuffer.wrap(buffer.getAllBytes(avoidCopying = false))
```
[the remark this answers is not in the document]
ac101m: On reflection, I actually think it makes sense to get rid of the byte buffers and replace them with plain byte arrays. They were used previously in the old handler system to take off muxing information before sending messages downstream, but this isn't necessary anymore. Thoughts?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveEnvironment.kt
```diff
@@ -19,33 +17,41 @@ import java.util.concurrent.atomic.AtomicLong
 
 @PotentialPackagePrivate
 class NativeEnclaveEnvironment(
-    enclaveClass: Class<*>
+    enclaveClass: Class<*>,
+    override val hostCallInterface: NativeHostCallInterface
```
bonifazi-r3, line 21: I am confused by this `hostCallInterface` member variable and the one in the companion object.
Are both of them needed?

bonifazi-r3 (changes requested):
Another couple of comments

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallAcceptor.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+abstract class CallAcceptor<CALL_ID_TYPE> {
+    private val callHandlers = ConcurrentHashMap<CALL_ID_TYPE, CallHandler>()
+
+    /**
+     * Add a call handler for a specific call type.
+     * If the call type already has a handler, throw an exception.
+     */
+    fun registerCallHandler(callType: CALL_ID_TYPE, handler: CallHandler) {
+        check(callHandlers[callType] == null) { "Call handler already registered for $callType." }
```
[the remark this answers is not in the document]
ac101m: Fixed in this case, however left as is in `acceptCall` as we'd need two map lookups otherwise.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInitiator.kt
```diff
@@ -0,0 +1,11 @@
+package com.r3.conclave.common.internal
+
+import java.nio.ByteBuffer
+
+interface CallInitiator<CALL_ID_TYPE> {
+    companion object {
+        private val EMPTY_BYTE_BUFFER: ByteBuffer = ByteBuffer.wrap(ByteArray(0)).asReadOnlyBuffer()
+    }
+
+    fun executeCall(callType: CALL_ID_TYPE, parameterBuffer: ByteBuffer = EMPTY_BYTE_BUFFER): ByteBuffer?
```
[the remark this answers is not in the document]
ac101m: It's just so that you can write `executeCall(<call type>)` without having to explicitly pass an empty byte buffer. It avoids a pointless object being created, that's all. I also considered making the parameter nullable. Thoughts?

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallAcceptor.kt
```diff
@@ -0,0 +1,29 @@
...
+        callHandlers[callType] = handler
+    }
+
+    /**
+     * Route a call to a registered call handler.
+     * If no handler is registered, throw an exception.
+     */
+    fun acceptCall(callType: CALL_ID_TYPE, parameterBuffer: ByteBuffer): ByteBuffer? {
```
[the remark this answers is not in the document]
ac101m: Thought about this actually, felt that `HostCallInterface : CallAcceptor<EnclaveCallType>, CallInitiator<HostCallType>` was less ambiguous than `HostCallInterface : CallMultiplexer<EnclaveCallType>, CallInitiator<HostCallType>`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/HostCallInterface.kt
```diff
@@ -0,0 +1,66 @@
+package com.r3.conclave.enclave.internal
+
+import com.r3.conclave.common.internal.*
+import com.r3.conclave.common.internal.attestation.Attestation
+import java.nio.ByteBuffer
+import java.security.KeyPair
+import java.security.PublicKey
+
+abstract class HostCallInterface : CallInitiator<HostCallType>, CallAcceptor<EnclaveCallType>() {
```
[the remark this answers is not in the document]
ac101m: How about just `EnclaveInterface` and `HostInterface`, would that resolve the ambiguity for you?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveEnvironment.kt
ac101m: They are the same instance. The `EnclaveEnvironment` class is our enclave-side abstraction for dealing with different enclave implementations (the counterpart to `EnclaveHandle` on the host side). For the call interface to be agnostic to the enclave implementation, the interface instance has to be in the `EnclaveEnvironment` instance (hence the override). In the case of the `NativeEnclaveEnvironment` the call interface is a singleton (in the companion) and is passed into the `NativeEnclaveEnvironment` when the object is constructed. In mock mode it works differently, and it will probably be different still when we implement the `GrameneEnclaveEnvironment` class.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeHostCallInterface.kt
ac101m: This is as intended. Null return values are signalled to the other side by not sending a return value or exception message. If the buffer is empty, then an ocall _will_ occur and an empty buffer will be sent. See `NativeHostCallinterface.kt:54`.

conclave-enclave/src/main/java/com/r3/conclave/enclave/internal/Native.java
```diff
@@ -10,6 +10,16 @@ public class Native {
      */
     public static native void jvmOcall(byte[] data);
 
+    // TODO: Temporary for con 1025
```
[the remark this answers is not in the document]
ac101m: I'm not sure I understand. You want me to add "con1025" back to all the method names? If so, I'm not doing that! :rofl:

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/EnclaveCallType.kt
```diff
@@ -0,0 +1,28 @@
...
+    START_ENCLAVE,
+    STOP_ENCLAVE,
+    GET_ENCLAVE_INSTANCE_INFO_QUOTE,
+    GET_KDS_PERSISTENCE_KEY_SPEC,
+    SET_KDS_PERSISTENCE_KEY,
+    SEND_MESSAGE_HANDLER_COMMAND;
+
+    fun toShort(): Short {
+        check(VALUES.size < Short.MAX_VALUE)
+        return ordinal.toShort()
```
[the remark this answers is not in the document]
ac101m: I felt at the time of writing that 127 call types might not be enough. In hindsight I think it's plenty and this isn't a public API so we can change it later if we have to. Changed to byte.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/EnclaveCallType.kt
```diff
@@ -0,0 +1,28 @@
...
+ * Enumeration of all possible host -> enclave calls
+ */
+enum class EnclaveCallType {
+    INITIALISE_ENCLAVE,
+    START_ENCLAVE,
+    STOP_ENCLAVE,
+    GET_ENCLAVE_INSTANCE_INFO_QUOTE,
+    GET_KDS_PERSISTENCE_KEY_SPEC,
+    SET_KDS_PERSISTENCE_KEY,
+    SEND_MESSAGE_HANDLER_COMMAND;
```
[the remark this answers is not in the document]
ac101m: I'm not sure either is good! The behaviour of the message handler is not strictly request/response (one of the reasons I didn't try to disentangle it's various functions). Renamed to just `CALL_MESSAGE_HANDLER`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/HostCallInterface.kt
```diff
@@ -0,0 +1,66 @@
...
+        }
+        return Cursor.slice(SgxSignedQuote, quoteBuffer)
+    }
+
+    /**
+     * Get quoting enclave info from the host.
+     */
+    fun getQuotingEnclaveInfo(): ByteCursor<SgxTargetInfo> {
+        val infoBuffer = checkNotNull(executeCall(HostCallType.GET_QUOTING_ENCLAVE_INFO)) {
+            MISSING_RETURN_VALUE_ERROR_MESSAGE
```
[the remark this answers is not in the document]
ac101m: Added `executeCallAndCheckReturn` to the call initiator interface

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -79,13 +78,15 @@ abstract class Enclave {
...
     private lateinit var signingKeyPair: KeyPair
-    private lateinit var adminHandler: AdminHandler
-    private lateinit var quotingEnclaveInfoHandler: QuotingEnclaveInfoHandler
-    private lateinit var signedQuoteHandler: SignedQuoteHandler
-    private lateinit var enclaveInstanceInfoQuoteHandler: EnclaveInstanceInfoQuoteHandler
-    private lateinit var enclaveMessageHandler: EnclaveMessageHandler
     private lateinit var aesPersistenceKey: ByteArray
 
+    private val startCallHandler = StartCallHandler()
+    private val stopCallHandler = StopCallHandler()
```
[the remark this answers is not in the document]
ac101m: Possibly, will check

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeHostCallInterface.kt
bonifazi-r3: As discussed, let's just add an explanation comment, feel free to close this once done

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInitiator.kt
```diff
@@ -0,0 +1,17 @@
+package com.r3.conclave.common.internal
+
+import java.nio.ByteBuffer
+
+interface CallInitiator<CALL_ID_TYPE> {
```
bonifazi-r3, line 5: This does not really seem a proper "old style" interface, not sure if we follow any convention though.
But here comes another bigger question: `CallInitiator`, together with `CallAcceptor` are always used together and we only have two combinations for them. We will never have more than 2 combinations I think.
I am questioning if we need this "templatization" ("genericsation", not sure how to call this in Java). 
Maybe we can at least merge them together in one class?

bonifazi-r3 (changes requested):

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/MockEnclaveHandle.kt
```diff
@@ -1,63 +1,60 @@
...
 
-    override val connection: CONNECTION = hostHandler.connect(this)
-
     override val enclaveClassName: String get() = mockEnclave.javaClass.name
 
-    private val enclaveHandler by lazy {
-        val sender = MockOcallSender(HandlerConnected(hostHandler, connection))
+    private val callInterfaceConnector = MockCallInterfaceConnector()
+
+    override val enclaveCallInterface = run {
```
bonifazi-r3, line 23: Isn't this called every time it is referenced?

bonifazi-r3 (changes requested):

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInitiator.kt
ac101m: I have combined these, it actually works quite well I think!

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/HostCallInterface.kt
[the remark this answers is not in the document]
ac101m: Changed to `EnclaveHostInterface` and `HostEnclaveInterface`

bonifazi-r3 (changes requested):
Minor comments, this is perfect now

bonifazi-r3 (commented):
This looks impressive and well structured work, it looks very good to me, thanks for your effort.
@ac101m ping me when you are ready for this to be approved.

conclave-common/src/main/kotlin/com/r3/conclave/common/MockCallInterfaceConnector.kt
```diff
@@ -0,0 +1,58 @@
+package com.r3.conclave.common
```
shamsasari, line 1: This is a public API package, and this class is internal. Move this to the internal package.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterface.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    private val callHandlers = ConcurrentHashMap<INCOMING_CALL_TYPE, CallHandler>()
+
+    /**
+     * Execute a call and maybe get a return buffer.
+     */
+    abstract fun initiateOutgoingCall(callType: OUTGOING_CALL_TYPE, parameterBuffer: ByteBuffer = EMPTY_BYTE_BUFFER): ByteBuffer?
+
+    /**
+     * Execute a call and get a return buffer. Throw an exception if no buffer is provided.
```
shamsasari, line 24: is returned

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterface.kt
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    /**
+     * Add a call handler for a specific call type.
+     * If the call type already has a handler, throw an exception.
+     */
+    fun registerCallHandler(callType: INCOMING_CALL_TYPE, handler: CallHandler) {
+        check(!callHandlers.containsKey(callType)) { "Call handler already registered for $callType." }
+        callHandlers[callType] = handler
```
shamsasari, line 38: Race condition!

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterface.kt
```diff
@@ -0,0 +1,49 @@
...
+    companion object {
+        private val EMPTY_BYTE_BUFFER: ByteBuffer = ByteBuffer.wrap(ByteArray(0)).asReadOnlyBuffer()
+    }
+
+    private val callHandlers = ConcurrentHashMap<INCOMING_CALL_TYPE, CallHandler>()
+
+    /**
+     * Execute a call and maybe get a return buffer.
+     */
+    abstract fun initiateOutgoingCall(callType: OUTGOING_CALL_TYPE, parameterBuffer: ByteBuffer = EMPTY_BYTE_BUFFER): ByteBuffer?
```
shamsasari, line 21: What does "initiate" mean? It implies the execution can be asynchronous, but I presume that's not the case. As you've documented, "execute" might be clearer.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterface.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Execute a call and get a return buffer. Throw an exception if no buffer is provided.
+     */
+    fun initiateOutgoingCallAndCheckReturn(callType: OUTGOING_CALL_TYPE, parameterBuffer: ByteBuffer = EMPTY_BYTE_BUFFER): ByteBuffer {
```
shamsasari, line 26: `AndCheckReturn` -> `withReturn`?

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterface.kt
shamsasari, line 21: Perhaps dataBuffer is clearer, especially for the `handleIncomingCall`

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterfaceMessageType.kt
```diff
@@ -0,0 +1,23 @@
+package com.r3.conclave.common.internal
+
+/**
+ * This forms part of the messaging protocol for enclave and host calls in native mode.
+ * Messages (ecalls & ocalls) may represent calls, returns or exceptions.
```
shamsasari, line 5: Let's use all uppercase for acronyms

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterfaceStackFrame.kt
```diff
@@ -0,0 +1,8 @@
+package com.r3.conclave.common.internal
+
+import java.nio.ByteBuffer
+
+class CallInterfaceStackFrame<CALL_TYPE>(
+        val callType: CALL_TYPE,
+        var returnBuffer: ByteBuffer? = null,
+        var exceptionBuffer: ByteBuffer? = null)
```
shamsasari, line 8: Minor: move the closing bracket to the next line

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/EnclaveCallType.kt
```diff
@@ -0,0 +1,28 @@
...
+ * Enumeration of all possible host -> enclave calls
+ */
+enum class EnclaveCallType {
+    INITIALISE_ENCLAVE,
+    START_ENCLAVE,
+    STOP_ENCLAVE,
+    GET_ENCLAVE_INSTANCE_INFO_QUOTE,
+    GET_KDS_PERSISTENCE_KEY_SPEC,
+    SET_KDS_PERSISTENCE_KEY,
+    CALL_MESSAGE_HANDLER;
```
shamsasari, line 13: This one seems quiet generic.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/EnclaveCallType.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    fun toByte(): Byte {
+        check(VALUES.size < Byte.MAX_VALUE)
```
shamsasari, line 16: A bit unnecessary I think. Better to document clearly instead.

conclave-enclave/src/main/java/com/r3/conclave/enclave/internal/Native.java
```diff
@@ -6,9 +6,12 @@
 public class Native {
     /**
      * Makes an OCALL.
+     * @param callTypeID When the host initiates a call, this contains the host call type. If the Ocall is a return or
```
shamsasari, line 9: OCALL

conclave-enclave/src/main/java/com/r3/conclave/enclave/internal/substratevm/EntryPoint.java
```diff
@@ -14,12 +15,12 @@
...
     @CEntryPoint(name = "Java_com_r3_conclave_enclave_internal_substratevm_EntryPoint_entryPoint")
-    public static void entryPoint(IsolateThread thread, CCharPointer input, int len) {
+    public static void entryPoint(IsolateThread thread, byte callTypeID, byte messageTypeID, CCharPointer data, int dataLengthBytes) {
         // As mentioned in the JavaDocs, the default byte order for a new ByteBuffer is _always_ big endian.
         // CTypeConversion.asByteBuffer however uses native byte order. So to make sure we don't break code that
         // assumes the default, we switch back to big endian.
-        ByteBuffer buffer = CTypeConversion.asByteBuffer(input, len).order(ByteOrder.BIG_ENDIAN).asReadOnlyBuffer();
-        NativeEnclaveEnvironment.enclaveEntry(buffer);
+        ByteBuffer parameterBuffer = CTypeConversion.asByteBuffer(data, dataLengthBytes).order(ByteOrder.BIG_ENDIAN).asReadOnlyBuffer();
+        NativeEnclaveEnvironment.enclaveEntry(callTypeID, CallInterfaceMessageType.Companion.fromByte(messageTypeID), parameterBuffer);
```
shamsasari, line 23: Annotate `fromByte` with `@JvmStatic` to avoid the `Companion` reference

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -150,7 +148,30 @@ abstract class Enclave {
...
+        // are insecure).
+        check(attestation.enclaveMode == env.enclaveMode) {
+            "The enclave mode of the attestation (${attestation.enclaveMode}) does not match ${env.enclaveMode}"
+        }
+
+        EnclaveInstanceInfoImpl(
+                signatureKey,
+                encryptionKeyPair.public as Curve25519PublicKey,
+                attestation
+        )
```
shamsasari, line 173: There's some security checks here and thus the entire lazy block is better served to be its own separate method.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -212,30 +233,20 @@ abstract class Enclave {
...
-            val mux = connected.connection.setDownstream(SimpleMuxingHandler())
-            adminHandler = mux.addDownstream(AdminHandler(this, env))
-            quotingEnclaveInfoHandler = mux.addDownstream(QuotingEnclaveInfoHandler())
-            signedQuoteHandler = mux.addDownstream(SignedQuoteHandler())
-            enclaveInstanceInfoQuoteHandler = mux.addDownstream(EnclaveInstanceInfoQuoteHandler())
-            enclaveMessageHandler = mux.addDownstream(EnclaveMessageHandler())
-            connected
+        initCryptography()
+
+        env.hostInterface.apply {
```
shamsasari, line 240: Can these be registered at creation time of the enclave object, i.e. inside an `init`?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/EnclaveHostInterface.kt
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Send enclave info to the host.
+     * TODO: It would be better to return enclave info from the initialise enclave call
+     *       but that doesn't work in mock mode at the moment.
+     */
+    fun setEnclaveInfo(signatureKey: PublicKey, encryptionKeyPair: KeyPair) {
+        val encodedSigningKey = signatureKey.encoded                    // 44 bytes
+        val encodedEncryptionKey = encryptionKeyPair.public.encoded     // 32 bytes
+        val payloadSize = encodedSigningKey.size + encodedEncryptionKey.size
+        val buffer = ByteBuffer.wrap(ByteArray(payloadSize)).apply {
```
shamsasari, line 19: allocate

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * This class is the implementation of the [EnclaveHostInterface] for native enclaves.
+ * It has three jobs:
+ *  - Serve as the endpoint for calls to make to the host, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Route calls from the host to the appropriate enclave side call handler, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Handle the low-level details of the messaging protocol (ecalls and ocalls).
+ */
+class NativeEnclaveHostInterface : EnclaveHostInterface() {
+    /** In release mode we want to sanitise exceptions to prevent leakage of information from the enclave */
+    var sanitiseExceptions: Boolean = false
```
shamsasari, line 21: Why can't this be a c'tor parameter?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
```diff
@@ -0,0 +1,121 @@
...
+     * When a message arrives from the host, this stack is used to associate the return value with the corresponding call.
+     */
+    private val threadStacks = ThreadLocal<Stack<StackFrame>>()
+    private val stack: Stack<StackFrame>
+        get() {
+            if (threadStacks.get() == null) {
+                threadStacks.set(Stack<StackFrame>())
+            }
+            return threadStacks.get()
+        }
```
shamsasari, line 34: This can be simplified if you can create your thread local using `withInitial` factory method.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
```diff
@@ -0,0 +1,121 @@
...
+ */
+class NativeEnclaveHostInterface : EnclaveHostInterface() {
+    /** In release mode we want to sanitise exceptions to prevent leakage of information from the enclave */
+    var sanitiseExceptions: Boolean = false
+
+    /**
+     * Each thread has a lazily created stack which contains a frame for the currently active host call.
+     * When a message arrives from the host, this stack is used to associate the return value with the corresponding call.
+     */
+    private val threadStacks = ThreadLocal<Stack<StackFrame>>()
```
shamsasari, line 27: Don't use `Stack` for stacks in Java! It's an unfortunate thing. All the methods are synchronized which adds an overhead. That overhead is now very small, but it's still better practice to use something like `ArrayDeque`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Handle call initiations from the host.
+     * This method propagates the call to the appropriate enclave side call handler. If a return value is produced or an
+     * exception occurs, a reply message is sent back to the host.
+     */
+    private fun handleCallEcall(callType: EnclaveCallType, parameterBuffer: ByteBuffer) {
+        try {
+            handleIncomingCall(callType, parameterBuffer)?.let {
```
shamsasari, line 91: Use a local variable rather than this multi-line `let` block.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/HostEnclaveInterface.kt
```diff
@@ -0,0 +1,76 @@
...
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+
+abstract class HostEnclaveInterface : CallInterface<EnclaveCallType, HostCallType>() {
+    /**
+     * Initialises the enclave by instantiating the specified class.
+     * This is not currently used in mock mode.
+     */
+    fun initializeEnclave(enclaveClassName: String) {
+        initiateOutgoingCall(EnclaveCallType.INITIALISE_ENCLAVE, ByteBuffer.wrap(enclaveClassName.toByteArray(StandardCharsets.UTF_8)))
```
shamsasari, line 17: UTF_8 is the default parameter value so no need to specify it.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/NativeHostEnclaveInterface.kt
```diff
@@ -0,0 +1,101 @@
...
+ *  - Serve as the endpoint for calls to make to the enclave, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Route calls from the enclave to the appropriate host side call handler, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Handle the low-level details of the messaging protocol (ecalls and ocalls).
+ */
+class NativeHostEnclaveInterface(private val enclaveId: Long) : HostEnclaveInterface() {
+    /**
+     * Each thread has a lazily created stack which contains a frame for the currently active enclave call.
+     * When a message arrives from the enclave, this stack is used to associate the return value with the corresponding call.
+     */
+    private val threadLocalStacks = ThreadLocal<Stack<StackFrame>>()
```
shamsasari, line 22: ditto comments

shamsasari (changes requested):
This is very well done and well thought out. Just some small changes needed.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/EnclaveCallType.kt
ac101m: So this is for the message handler, which has been left as is (more or less). The messages for the message handler are sometimes commands, sometimes responses, so the name ended up being kinda generic. Disentangling the message handlers individual functions was taking me too long to figure out, so I decided I'd leave it alone for now and raise a separate ticket to deal with it later.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
ac101m: It's not know at construction time if we're in release mode or not.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/EnclaveCallType.kt
ac101m: I've removed the runtime check and added a unit test just to be thorough.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
ac101m: Possibly, but not easily.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterface.kt
```diff
@@ -33,6 +33,7 @@ abstract class CallInterface<OUTGOING_CALL_TYPE, INCOMING_CALL_TYPE> {
      * Add a call handler for a specific call type.
      * If the call type already has a handler, throw an exception.
      */
+    @Synchronized
```
shamsasari, line 36: You're using `ConcurrentHashMap`, so just make use of `putIfAbsent` to avoid locking as much as possible.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterfaceStackFrame.kt
```diff
@@ -0,0 +1,9 @@
+package com.r3.conclave.common.internal
+
+import java.nio.ByteBuffer
+
+class CallInterfaceStackFrame<CALL_TYPE>(
```
shamsasari, line 5: Can you add a KDoc for this class, since it's non-trivial what this class is involved in.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
```diff
@@ -0,0 +1,120 @@
...
+    private val threadLocalStacks = ThreadLocal.withInitial { ArrayDeque<StackFrame>() }
+    private val stack get() = threadLocalStacks.get()
+
+    private fun checkCallType(type: HostCallType) = check(type == stack.last().callType) { "Call type mismatch" }
+
+    /**
+     * Internal method for initiating a host call with specific arguments.
+     * This should not be called directly, but instead by implementations in [EnclaveHostInterface].
+     */
+    override fun executeOutgoingCall(callType: HostCallType, parameterBuffer: ByteBuffer): ByteBuffer? {
```
shamsasari, line 36: `executeOCall` would be even more clearer

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
```diff
@@ -0,0 +1,120 @@
...
+        stack.addLast(StackFrame(callType, null, null))
+
+        Native.jvmOCall(
+                callType.toByte(), CallInterfaceMessageType.CALL.toByte(), parameterBuffer.getAllBytes(avoidCopying = true))
+
+        val stackFrame = stack.removeLast()
```
shamsasari, line 42: Is `stackFrame` the same as the one that was added at the beginning of the method? Should we have an instance equality (`===`) check for sanity?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
ac101m: That would confuse this with edl OCalls, which this isn't

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/CallInterface.kt
ac101m: How will the error condition be detected in that case?

shamsasari: Have a look at the javadoc - it returns null if there is no previous mapping

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/NativeEnclaveHostInterface.kt
ac101m: It is, or at least it should be. Added just to ensure nothing goes awry.

shamsasari (approved):
Great work!

@bonifazi-r3 if you could also take another look so that we can get this merged.

bonifazi-r3 (approved):

PR #13, opened 2022-08-25
CON-1020: Publish Conclave SDK to Maven Central (fix publication to all repositories)

bonifazi-r3:
This is an add-on to the previous PR https://github.com/R3Conclave/conclave-core-sdk/pull/9, as we missed to specify in the script that we want to publish to Maven Central repo.
We are now publishing to all repositories, including Maven Central


shamsasari (approved):
Thanks

PR #14, opened 2022-08-25
Updated conclave-init generated project to not use deprecated Gradle API

shamsasari:
This stops ugly deprecation warning from being displayed when the project is built.


filipesoliveira (approved):

shamsasari:
I've intentionally put this in the release/1.3 branch. My thinking being if we have a 1.3.1 release then it will get this fix for free. Otherwise it will land in 1.4 on the next release/1.3 -> master merge.

shamsasari:
And also because this is a simple fix.

PR #16, opened 2022-08-26
CON-1020: Publish Conclave SDK to Maven Central (address Sonatype validation issues)

bonifazi-r3:
This patch solves validation issues reported while moving the artifacts from staging to release repository in Sonatype

### DIFF
--- a/plugin-enclave-gradle/build.gradle
+++ b/plugin-enclave-gradle/build.gradle
@@ -1,10 +1,16 @@
-//  Note: the settings specifying the mapping of this plugin
-//    to its implementation class are in the root build.gradle
-
 plugins {
+    id 'java-gradle-plugin'
     id 'org.jetbrains.kotlin.jvm'
 }
 
+gradlePlugin {
+    plugins {
+        enclavePlugin {
+            id = 'com.r3.conclave.enclave'
+            implementationClass = 'com.r3.conclave.plugin.enclave.gradle.GradleEnclavePlugin'
+        }
+    }
+}
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
--- a/plugin-enclave-gradle/build.gradle
+++ b/plugin-enclave-gradle/build.gradle
@@ -1,16 +1,10 @@
+//  Note: the settings specifying the mapping of this plugin
+//    with its implementation class are in the root build.gradle
+
 plugins {
-    id 'java-gradle-plugin'
     id 'org.jetbrains.kotlin.jvm'
 }
 
-gradlePlugin {
-    plugins {
-        enclavePlugin {
-            id = 'com.r3.conclave.enclave'
-            implementationClass = 'com.r3.conclave.plugin.enclave.gradle.GradleEnclavePlugin'
-        }
-    }
-}
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
--- a/plugin-enclave-gradle/build.gradle
+++ b/plugin-enclave-gradle/build.gradle
@@ -150,11 +150,10 @@ processResources {
         from(generateCAPCache)
     }
 }
-/*
+
 jar {
     manifest {
         // This is used in the plugin to add implicit dependencies of the right version.
         attributes("Conclave-Version": rootProject.version.toString())
     }
 }
-*/
--- a/plugin-enclave-gradle/build.gradle
+++ b/plugin-enclave-gradle/build.gradle
@@ -4,8 +4,6 @@ plugins {
 }
 
 gradlePlugin {
-    automatedPublishing = false
-
     plugins {
         enclavePlugin {
             id = 'com.r3.conclave.enclave'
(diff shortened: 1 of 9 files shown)


plugin-enclave-gradle/build.gradle
```diff
@@ -4,6 +4,8 @@ plugins {
 }
 
 gradlePlugin {
+    automatedPublishing = false
```
bonifazi-r3, line 7: This is to avoid a double publication of the artifact of this project. The last one (`pluginMaven`) is overriding the first one, which is created in the main `build.gradle` file

test-conclave-init.sh
```diff
@@ -11,11 +11,11 @@ echo
 
 pushd build
 
-conclaveInitJar=$(find repo/com/r3/conclave/conclave-init/ -name 'conclave-init-*.jar')
+conclaveInitJar=$(find repo/com/r3/conclave/conclave-init/ find -name 'conclave-init-*jar' -not -name 'conclave-init-*javadoc.jar' -not -name 'conclave-init-*-sources.jar')
```
bonifazi-r3, line 14: As we are generating javadocs and sources now, this script needs more filtering, to actually find, for example,  `conclave-init-1.3.jar`

test-conclave-init.sh
```diff
@@ -11,11 +11,11 @@ echo
 
 pushd build
 
-conclaveInitJar=$(find repo/com/r3/conclave/conclave-init/ -name 'conclave-init-*.jar')
+conclaveInitJar=$(find repo/com/r3/conclave/conclave-init/ -name 'conclave-init-*jar' -not -name 'conclave-init-*javadoc.jar' -not -name 'conclave-init-*-sources.jar')
```
bonifazi-r3, line 14: As we are generating javadocs and sources now, this script needs more filtering, to actually find, for example, conclave-init-1.3.jar

shamsasari: If we read in the conclave version from versions.gradle then we can be exact about the file name we want, and thus not need narrow done manually.

bonifazi-r3: I prefer not to do that to be honest, as it will create a dependency somehow that we do not need. I.e. I see this script as something that in theory we can run with the repo only, without the version file

plugin-enclave-gradle/build.gradle
```diff
@@ -1,16 +1,10 @@
...
 }
 
-gradlePlugin {
-    plugins {
-        enclavePlugin {
-            id = 'com.r3.conclave.enclave'
-            implementationClass = 'com.r3.conclave.plugin.enclave.gradle.GradleEnclavePlugin'
-        }
-    }
-}
```
bonifazi-r3, line 8: I have moved this to the root `build.gradle` as `java-gradle-plugin` works with `maven-publish` plugin during publication and this issues if we do not coordinate the two together in the same place (i.e. for a developer it is a pain to discover that a change in `build.gradle` is also affected by the `java-gradle-plugin` here. This is in the context of the evaluation of the project.

bonifazi-r3:
@shamsasari This is currently generating the pom files correctly. If this now  passes the tests, we can try again to publish and see what happens, hopefully I haven't forgotten anything

plugin-enclave-gradle/build.gradle
```diff
@@ -1,16 +1,10 @@
+//  Note: the settings specifying the mapping of this plugin
+//    to its implementation class are in the root build.gradle
+
 plugins {
-    id 'java-gradle-plugin'
     id 'org.jetbrains.kotlin.jvm'
 }
 
-gradlePlugin {
```
shamsasari, line 6: Is it necessary to move this? This configuring this module to be a Gradle plugin, which is a property independent of publishing, and thus is best placed here.

plugin-enclave-gradle/build.gradle
shamsasari: Can this not be resolved by just commenting in the root build.gradle. See my comment above why I'd prefer this config to stay here.

plugin-enclave-gradle/build.gradle
bonifazi-r3: Not strictly, but it needs `java-gradle-plugin` which I moved to avoid collateral issues as `java-gradle-plugin` interacts with `maven-publish` to generate the publications for the plugin.

If we want to keep this here, I would still need `java-gradle-plugin` in the root `build.gradle`. And I would need to put a comement here

shamsasari: >If we want to keep this here, I would still need java-gradle-plugin in the root build.gradle

I think that's OK.

bonifazi-r3: Let me put it back, will write the comment

plugin-enclave-gradle/build.gradle
bonifazi-r3: Moved back

plugin-enclave-gradle/build.gradle
bonifazi-r3: Should be fine now

shamsasari (approved):

PR #17, opened 2022-08-26
CON-394: Using JElf library to read enclave metadata in integration tests

shamsasari:
This allows us to remove some native code, which is otherwise not used.


shamsasari:
The failing test seems to be caused by https://github.com/intel/linux-sgx/issues/360.

PR #20, opened 2022-08-29
CON-1125: Update tutorials

JohnJoser3:
Updated 'running-hello-world' and 'writing-hello-world' docs.

- Changes for open sourcing.
- Updates as per content audit.
- Rewrite for simplicity and clarity.
- Reduce information overload.


docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
 
-This would generate the host JAR at `host/build/libs/host-simulation.jar`.
+!!!Note
 
-!!! tip
-    If working from inside IntelliJ, find the bootJar task of your host project from the Gradle menu on the right-hand 
-    side. Modify run configuration of this task by adding the -PenclaveMode=<MODE> flag to the arguments section.
+    If you use IntelliJ, find the bootJar task of your host project from the Gradle menu on the right-hand 
+    side. Modify the run configuration of this task by adding the -PenclaveMode=<MODE> flag to the arguments section.
     Now you can run the task.
```
shamsasari, line 183: I don't think we should be providing such guidance on Intellij. But I leave it up to you.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
 
-For **release mode**, the sample app has been configured (in the `build.gradle` of the `enclave` subproject) to use
-external signing. Note that **external signing is optional** and you do not have to include it in your own projects.
-See [Enclave signing](signing.md) for more information on external signing.
+For *release mode*, the sample app is configured (in the `build.gradle` of the `enclave` subproject) to use
```
shamsasari, line 185: Perhaps "external code signing" to more clear, in line with my previous comment.

To give a quick bit of background. There's two types of "signing" that's linked to an enclave:
* This "code signer" or "code signing", which is the enclave's binary file (which is loaded by the host application) being signed by an external RSA key.
* Inside the enclave code, the enclave has the ability to sign arbitrary data bytes for the user. This is what what "signing" would probably normally mean to a developer.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
 === "macOS"
-    A Linux environment is required to run the host in simulation mode. This section describes how to instantiate such an
-    environment using Conclave's Docker integration.
 
-    !!! note
-        * Docker is only required to run the *host* on macOS. The client can be run without Docker.
+You need a Linux environment to run the host in simulation mode. This section describes how to create such an
+environment using Conclave's Docker integration.
+
+    !!!Note
```
shamsasari, line 262: For macOS we need to mention that Apple Silicon isn't supported

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
+Follow the instructions to run the app in [mock mode](enclave-modes.md). After running it in mock mode, you
+can proceed to [run the app in other modes](#beyond-mock-mode).
 
-If you get stuck, join our [discord server](https://discord.com/invite/dDaBGqGPre) and ask for help!
+If you get stuck at any step, please [talk to us on Discord](https://discord.gg/zpHKkMZ8Sw).
 
 ## Prerequisites
 
-* You need JDK 17 installed in order to build and run the app.
 * This tutorial assumes you've read and understood the [conceptual overview](enclaves.md)
```
shamsasari, line 16: enclaves.md is actually a big page, and it doesn't make sense to make it a prerequisite of running hello world. It defeats the whole purpose of trying to get a new user comfortable with Conclave. I would actually say this page would be further reading after they've run the sample.

docs/docs/writing-hello-world.md
```diff
@@ -1,45 +1,47 @@
...
 
 ## Project Setup
-Start by creating a new Conclave project using [Conclave Init](conclave-init.md).
+
+To set up a project:
+
+1. Download the Conclave Init tool.
 ```bash
-java -jar <path to sdk>/tools/conclave-init.jar \
+   wget https://repo1.maven.org/maven2/com/r3/conclave/conclave-init/1.4/conclave-init-1.4.jar -O conclave-init.jar
```
shamsasari, line 25: First of all, the URL needs to be pointing to 1.3, not 1.4, as that is what the docsite will go live on.

Secondly, I would rather that we didn't duplicate the download URL from conclave-init.md. The reason for this is that this URL needs to be manually updated as part of the release process. Would it be possible to say,

`follow the instructions to download Conclave [here](conclave-init.md)` and link to that page? 

However I understand that it is better user experience to not have to tell the user to get somewhere else to download the tool, and then come back and continue with the instructions. So perhaps it's fine to have the URL here as well. We can also update the release process instructions to update the URL on two pages!

docs/docs/writing-hello-world.md
```diff
@@ -77,30 +79,34 @@ each of the different build types.
 
 * Simulation builds use the default dummy key.
 * Debug builds use a private key stored in a file.
-* Release builds use a private key managed by some external signing process.
+* Release builds use a private key managed by an external signing process.
+
+2. Copy the `signing` directory in the `hello world`
```
shamsasari, line 84: OMG, we should not be telling users to copy a sample private key!!

The instructions here will need to updated to tell the user how to generate their own signing key.

docs/docs/writing-hello-world.md
```diff
@@ -77,30 +79,34 @@ each of the different build types.
...
-Alternatively you can provide or [generate your own](signing.md#generating-keys-for-signing-an-enclave) keys.
+Alternatively, you can also [generate your own](signing.md#generating-keys-for-signing-an-enclave) keys.
 
-!!! important
-    These keys aren't whitelisted by Intel so you can't use them for real release builds.
-    Only use these sample keys for the tutorial. Don't use them for signing your own enclaves!
+!!!Important
+
+    Use these sample keys only for the tutorial. Do not use these keys to sign your own enclaves as Intel hasn't
+    whitelisted these sample keys.
```
shamsasari, line 101: The whole concept of whitelisting signing keys is now dropped by Intel. Delete this "important" box. It's wrong also because we shouldn't be guiding developers to copy a publically visible sample private key!

docs/docs/writing-hello-world.md
```diff
@@ -135,79 +141,46 @@ public class ReverseEnclave extends Enclave {
...
 
-This method is called by the host when messages are sent to the enclave and makes use of the
-[Conclave mail](architecture.md#mail) API to facilitate encryption and authentication of messages. In the case of the
-enclave in this tutorial, it is also where replies are sent back to the host using the
-[`postMail`](api/-conclave%20-core/com.r3.conclave.enclave/-enclave/post-mail.html) method. The host then delivers the
-encrypted reply back to the appropriate client. In this tutorial, we will be handling transport using the built in
-[conclave web host](conclave-web-host.md), which allows clients to deliver and receive mail items via a simple REST API.
-In practice however, any method of transport can be used (see [writing your own host](writing-your-own-enclave-host.md)
-for an example using plain TCP sockets).
+The host calls this method to send messages to the enclave and uses the [Conclave mail](architecture.md#mail)
```
shamsasari, line 157: Mail

docs/docs/writing-hello-world.md
```diff
@@ -223,37 +196,36 @@ Assessed security level at 2020-07-17T16:31:51.894697Z is INSECURE
...
-actually considered secure yet because we're running in simulation mode still. An enclave can be `SECURE`, `STALE`,
-or `INSECURE`. An assessment of `STALE` means there is a software/firmware/microcode update available for the platform
-that improves security in some way. The client may wish to observe when this starts being reported and define a
-time span in which the remote enclave operator must upgrade.
-
-An attestation doesn't inherently expire but because the SGX ecosystem is always moving, client code will typically have
-some frequency with which it expects the host code to refresh the
-[`EnclaveInstanceInfo`](api/-conclave%20-core/com.r3.conclave.common/-enclave-instance-info/index.html). At present this is
-done by stopping/closing and then restarting the enclave.
+The hash in the first line is the *measurement*. This is a hash of the code of the enclave. It includes all
```
shamsasari, line 199: I feel this entire section about the remote attestation and EnclaveInstanceInfo belongs in its one section, not here about writing the hello world. Do we have a page where this would be better placed? Dp you know @anien or @bonifazi-r3 ?

docs/docs/writing-hello-world.md
```diff
@@ -315,45 +274,29 @@ public class ReverseEnclaveClient {
...
 
-The `constraint`, provided to the
-[`EnclaveClient` constructor](api/-conclave%20-core/com.r3.conclave.client/-enclave-client/-enclave-client.html) is the same
-constraint that we supplied when running the client
-in [Running your first enclave](running-hello-world.md#run-the-client)
+To encrypt the response from the enclave to the client, the client needs a public key and a private key.
 
-The client wants to receive a response from the enclave, and we want that to be encrypted/tamperproofed too. That means it
-needs a key pair of its own.
+Conclave abstracts the complexity of dealing with private keys inside `EnclaveClient`. The
```
shamsasari, line 284: Add back the links to the API docs

docs/docs/writing-hello-world.md
```diff
@@ -464,42 +416,30 @@ public class NativeTest {
...
-You'll notice that we annotated the test class with `@EnabledOnOs(OS.LINUX)`. This is from
-[JUnit 5](https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-os) and it will make sure
-the native test isn't run on non-Linux environments.
-
-The tests can use any enclave mode: release, debug, simulation or mock. Therefore, if you want to run your tests on
-a non-Linux system then you can configure your tests to depend on a mock mode enclave instead. In this case, remove the
-`@EnabledOnOs(OS.LINUX)` annotation from the above code.
+The test class is annotated with `@EnabledOnOs(OS.LINUX)`. This is from
+[JUnit 5](https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-os) to ensure that the
+native test is run on a Linux environment.
```
shamsasari, line 421: is only run

shamsasari (changes requested):

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
 
 ## Compile the sample application
 
-Clone and navigate to the hello world sample.
+Like all Conclave apps, this sample app has an [enclave and a client](architecture.md#primary-entities). The enclave
+runs inside a host, which is also a part of the sample app.
+
+To get the sample app and compile it:
+
+1. Open a command line interface.
```
anien, line 27: I don't think we need to tell users to open a command line interface. We don't mention it anywhere else in our docs before the bash commands. I would just remove this line.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+    Setting up an HTTPS connection is beyond the scope of this tutorial. You can try configuring Spring Boot or 
+    setting up a reverse proxy such as Nginx or Apache. See [Conclave web host](conclave-web-host.md) for more 
+    information.
 
 ## Run the client
 
-Run the client with
+After running the host:
+
+1. Open another command line interface.
```
anien, line 102: Similarly as above, I would remove first step.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-  # The string to reverse
-  reverse-me
+cd conclave-tutorials/hello-world
 ```
+3. Run the client with the command:
+```bash
+java -jar client/build/libs/client.jar "S:0000000000000000000000000000000000000000000000000000000000000000 PROD:1
+SEC:INSECURE" reverse-me
+```
+In the above command, the parameters in quotes are the constraints, and 'reverse-me' is the string to be reversed.
```
anien, line 112: We lost comment about constraint here, so I would add a link on the word constraints to the next section which explains it a bit more.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-  # A constraint which defines which enclaves the client will accept.
-  "S:0000000000000000000000000000000000000000000000000000000000000000 PROD:1 SEC:INSECURE" \
-  # The string to reverse
-  reverse-me
+cd conclave-tutorials/hello-world
 ```
+3. Run the client with the command:
+```bash
+java -jar client/build/libs/client.jar "S:0000000000000000000000000000000000000000000000000000000000000000 PROD:1
+SEC:INSECURE" reverse-me
```
anien, line 110: When rendered and copied to the clipboard and pasted to the command line, it is split in two lines and user has to manually delete the space to make the command work. I've seen this few times and it's annoying.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-   enclave was running in the expected mode.
-4. The client encrypted the "reverse-me" string using the public key provided in the attestation.
-5. As part of the message to the enclave, the client also included a public key for the enclave to use when responding.
-6. The client sent the encrypted message to the host's REST API.
-7. The host passed the encrypted message to the enclave.
-8. The enclave decrypted the client's message, reversed the string, and encrypted the response using the client's
+
+You have run your first enclave successfully.
+
+The following sequence of events produces the above output:
```
anien, line 121: This is very important explanation, I would highlight the 'title' somehow.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-9. The enclave passed the encrypted message to the host, was sent back to the client as an HTTP response to the same
+9. The enclave passes the encrypted message to the host, and is sent back to the client as an HTTP response to the same
    request.
 
-Congratulations on running your first enclave!
-
 ### Enclave Constraint
-The `--constraint` parameter of the client command-line interface defines the 
-properties of an acceptable enclave. This includes the signing key 
+The `--constraint` parameter of the client defines the properties of an acceptable enclave.
```
anien, line 137: We actually don't use `--constraint` in this example anymore. I would just say _The constraint parameter of the client_

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-Try rerunning the command with a different code signer:
+Try rerunning the command with a different signing key:
 ```
-"S:2222222222222222222222222222222222222222222222222222222222222222 PROD:1 SEC:INSECURE"
+java -jar client/build/libs/client.jar "S:2222222222222222222222222222222222222222222222222222222222222222 PROD:1
+SEC:INSECURE" reverse-me
 ```
 
-You should see the following error:
+You will get the following error, as the client's signing key(code signer) does not match the enclave's signing key:
```
anien, line 149: space before the opening parenthesis

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
 ```
-which we saw in the host output in the previous
-section.
 
-Try rerunning the command with a different code signer:
+Try rerunning the command with a different signing key:
 ```
-"S:2222222222222222222222222222222222222222222222222222222222222222 PROD:1 SEC:INSECURE"
+java -jar client/build/libs/client.jar "S:2222222222222222222222222222222222222222222222222222222222222222 PROD:1
+SEC:INSECURE" reverse-me
```
anien, line 146: Same problem with pasting the command to the command line.

docs/docs/running-hello-world.md
anien: Same thinking here. We used to have screenshots of how to build hello world in IntelliJ and in that case this note made sense. But as we now show all the commands in the command line, this doesn't really make sense/doesn't suit here.

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
 
-If you get stuck, join our [discord server](https://discord.com/invite/dDaBGqGPre) and ask for help!
+If you get stuck at any step, please [talk to us on Discord](https://discord.gg/zpHKkMZ8Sw).
 
 ## Prerequisites
 
-* You need JDK 17 installed in order to build and run the app.
 * This tutorial assumes you've read and understood the [conceptual overview](enclaves.md)
   and [architecture overview](architecture.md).
+* You need JDK 17 or higher installed. You can download it [here](https://www.oracle.com/java/technologies/downloads/).
```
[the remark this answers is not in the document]
bonifazi-r3: Do we really support JDK8? I would not mention it anyway, do we want the user to use it?

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
-# Compiling and Running your First Enclave
+# Compiling and running your first enclave
 
 ## Introduction
 
-This tutorial describes how to compile and run the sample hello world app, which you can find in the 
-[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world). This app contains the `ReverseEnclave`, 
-which reverses a string provided by the client and passes it back.
+This tutorial describes how to compile and run the sample hello world app, which you can find in the
+[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world).
```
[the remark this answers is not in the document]
bonifazi-r3: Maybe it is even better to link "sample hello world app" and remove `which you can find...`.
The repo of the app is just below

docs/docs/running-hello-world.md
bonifazi-r3: I agree. I would even remove the whole thing. As a user I would really want to avoid verbose words and get to the meat of getting the app up and running (and then, only after that, I would want to understand details

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
shamsasari: We do support JDK 8 (or at least we should do) @anien I presume we test this?

But yes, we drop the mention of JDK 8 and just day "upto JDK 17 is supported". But also add that JDK 17 is recommended as well.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
   - Product ID: 1
   - Revocation level: 0
 ```
-This is the [remote attestation](enclaves.md#remote-attestation), an object which proves certain information about the
-enclave that has been loaded. The private key that corresponds to the "Session encryption key" (printed above) is only
-available inside the enclave. Our client will use the "Session encryption key" to encrypt data to send into the enclave.
+This output is the [remote attestation](enclaves.md#remote-attestation), an object which proves certain information
+about the enclave. The private key corresponding to the "Session encryption key" (printed above) is only available
+inside the enclave. The client will use the "Session encryption key" to encrypt data to send to the enclave.
```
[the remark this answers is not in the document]
bonifazi-r3: Should we mention that it is the corresponding "public" part of the key that gets used by the client?

docs/docs/running-hello-world.md
bonifazi-r3: I agree with @anien , this deserves a title.
I would also personally make emphasis on the "entities" (host/client/enclave). Not sure if it's a good idea to have them upper cases or make them bold.
As this is a "software use case" one convention is to have entities underlined and the action in bold.
I.e. something like

The <ins>host</ins> **loads** the <ins>enclave</ins>, which **generates** a remote attestation report.

Not sure if this is nice to see though, you're the expert :-)

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-7. The host passed the encrypted message to the enclave.
-8. The enclave decrypted the client's message, reversed the string, and encrypted the response using the client's
+
+You have run your first enclave successfully.
+
+The following sequence of events produces the above output:
+
+1. The host loads the enclave, which generates a remote attestation report.
+2. The client connects to the host and downloads the attestation report.
+3. The client uses the attestation report to verify that it is connected to the enclave it expected, and that the
```
bonifazi-r3, line 125: maybe  `to the expected enclave`

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+   enclave is running in the expected mode.
```
bonifazi-r3, line 126: I would not make any reference to the modes here, I would make things as simple as possible

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+4. The client encrypts the "reverse-me" string using the public key provided in the attestation.
+5. In the message to the enclave, the client also includes a public key for the enclave to use when responding.
+6. The client sends the encrypted message to the host's REST API.
+7. The host passes the encrypted message to the enclave.
+8. The enclave decrypts the client's message, reverses the string, and encrypts the response using the client's
    public key.
-9. The enclave passed the encrypted message to the host, was sent back to the client as an HTTP response to the same
+9. The enclave passes the encrypted message to the host, and is sent back to the client as an HTTP response to the same
```
bonifazi-r3, line 133: "is sent back" passive form! :-)

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
    request.
 
-Congratulations on running your first enclave!
-
 ### Enclave Constraint
```
bonifazi-r3, line 136: I am wondering if the whole section is needed here. Maybe at the end? Or a link to a better explanation?

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
anien: @shamsasari , @bonifazi-r3 
We do support JDK 8 in Conclave and we do test it. BUT hello world requires Java 17. It doesn't work with any other JDK (JDK8, JDK11 or JDK18).
So here we need to say JDK 17 is required.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-when you are not using mock mode, which will be reflected in the Enclave Constraint provided to the client.
+when you are not using mock mode, which will be reflected in the enclave constraint provided to the client.
 
-When you run the host, you should see the new signing key in the output: 
+1. When you run the host, you should see the new signing key in the output:
 ```bash hl_lines="3"
-[main] INFO com.r3.conclave.host.web.EnclaveWebController - Remote attestation for enclave A92F481B7EEAE42D3EBB162BF77613605AF214D77D2E63D75A610FD485CFD7D6:
+[main] INFO com.r3.conclave.host.web.EnclaveWebController -
+Remote attestation for enclave A92F481B7EEAE42D3EBB162BF77613605AF214D77D2E63D75A610FD485CFD7D6:
   - Mode: MOCK
```
anien, line 307: We are saying above that we are running in different modes than MOCK and here we mention MOCK which is not correct. Also as code signer is different, this shouldn't be mock mode, but simulation or debug

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
 git clone https://github.com/R3Conclave/conclave-tutorials.git
 cd conclave-tutorials/hello-world
 ```
+3. Generate a fat JAR for the host and client using the `:bootJar` and `:shadowJar` tasks.
 
-The sample app, like all Conclave apps, consists of an
-[enclave and a client](architecture.md#primary-entities). The enclave runs inside a host application,
-which is provided by the Conclave Core SDK. We can generate a fat JAR for the host and client using the `:bootJar` and
-`:shadowJar` tasks, respectively.
 === "Windows"
```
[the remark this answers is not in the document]
anien: Maybe this needs a broader discussion. I personally like we do show users how to run hello world on all OSs. That way everyone can simply try our hello world. To your previous point Marco, I actually agree that for this tutorial, we should keep it very simple. Just something that user runs. They can dig into modes etc later if they want to. Maybe this tutorial can be simple steps to just run the hello world in mock mode (without explaining mock mode). How to run in different modes, explain the modes, use external signing for release mode can be explained on a different page.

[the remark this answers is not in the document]
bonifazi-r3: Yes, removing Windows/Mac is too extreme, you are right. What I personally dislike is the continuous conditionality of the tutorials. I.e. I'd prefer to have this tutorial for Linux, then for Windows, then for Mac. Not all together and interleaved/mixed.
Also, I would really put all the explanations at the end, in a separate paragraph.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
 
-For more information on how to choose a constraint, see [Enclave constraints](constraints.md).
-
+For more information on choosing a constraint, see [Enclave constraints](constraints.md).
 
 ## Beyond mock mode
-Eventually you will want to run your application in one of the other [enclave modes](enclave-modes.md), namely simulation,
-debug, or release. This section describes how to build and run the sample in these other modes.
+This section describes how to build and run the sample in other [enclave modes](enclave-modes.md), namely
+simulation, debug, and release.
```
anien, line 158: I think this is the first time we do mention enclave modes. Mode _debug_ is not used in the rest of the doc. I would add it along the simulation one to highlight it a bit more - see suggestion below (or as I commented above, maybe we can completely move these sections to a new doc).

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
     gradlew.bat :host:bootJar -PenclaveMode=simulation
-    ```
+```
 === "macOS / Linux"
-    ```bash
+```bash
     ./gradlew :host:bootJar -PenclaveMode=simulation
-    ```
+```
+
```
anien, line 176: Maybe to add a sentence: _Replace the simulation with debug for a debug mode._

docs/docs/running-hello-world.md
anien: I would highlight more release mode, maybe even as a new sub heading

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
anien: I don't agree to have the tutorial for Linux first and then for Windows/Mac. If you just simply switch the tabs depending on your OS for the commands, it seems nice and tidy solution to me. And the default tab should be Linux.

docs/docs/running-hello-world.md
JohnJoser3: I understand your angle. 

However, not every developer who reads this tutorial will know they must open two command lines - one for the host and one for the client. That's why I have added this step only in the tutorial. In most other cases, we don't want to tell this explicitly. But here, it will help a section of the readers.

docs/docs/running-hello-world.md
JohnJoser3: Same comment as above.

anien: And what about: _After running the host, in another command line interface:_ 
So then we don't need to mention it in the host section.

JohnJoser3: Correct. I will remove it.

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
 === "macOS / Linux"
-    ```bash
+```bash
     ./gradlew :host:bootJar :client:shadowJar
-    ```
-
-!!! note
-    * At this time using the JPMS tool `jlink` is not tested.
+```
+When the compilation is complete, you will get a success message: 'BUILD SUCCESSFUL'.
```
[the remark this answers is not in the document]
JohnJoser3: This is the output I get after compilation:

``` Starting a Gradle Daemon (subsequent builds will be faster)

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.3.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 16s
9 actionable tasks: 2 executed, 7 up-to-date ```

Will this always be the same? How much of this output should be in the tutorial?

File: docs/docs/running-hello-world.md [hunk 32]
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+[main] c.r.c.host.web.EnclaveWebHost$Companion  : Started EnclaveWebHost.Companion in <SECONDS> seconds 
 ```
 
-!!! warning
-    The sample uses Conclave mail which handles encryption and authentication, but the web host protocol makes use of
-    request headers which are not encrypted. Because of this, it is still important to use HTTPS for anything other than
-    internal development. Setting up an HTTPS connection is beyond the scope of this tutorial, but you may wish to look
-    at configuring Spring Boot or setting up a reverse proxy such as Nginx or Apache. See
-    [Conclave web host](conclave-web-host.md) for more information.
+!!!Warning
```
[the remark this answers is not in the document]
JohnJoser3: Do you propose to delete the entire 'Warning' call-out box? I too think it's good to remove it entirely.

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
JohnJoser3: Linked sample hello world app.

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
+# Compiling and running your first enclave
 
 ## Introduction
 
-This tutorial describes how to compile and run the sample hello world app, which you can find in the 
-[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world). This app contains the `ReverseEnclave`, 
-which reverses a string provided by the client and passes it back.
+This tutorial describes how to compile and run the sample hello world app, which you can find in the
+[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world).
+This app contains the `ReverseEnclave` function, which reverses a string provided by the client and passes it back.
```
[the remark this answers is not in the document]
JohnJoser3: Done.

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
 
-We are first going to run the app in [mock mode](enclave-modes.md),
-which is the simplest (and least realistic) of the enclave modes.
-We will explore running in other modes [later on](#beyond-mock-mode).
+Follow the instructions to run the app in [mock mode](enclave-modes.md). After running it in mock mode, you
+can proceed to [run the app in other modes](#beyond-mock-mode).
```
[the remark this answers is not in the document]
JohnJoser3: Shifted the references to different modes to a later section.

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
JohnJoser3: Rewrote that JDK 17 is required.

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Removed.

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
JohnJoser3: Removed the section to declutter the tutorial as part of another comment.

docs/docs/running-hello-world.md
JohnJoser3: Added an H4 subheading.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Removed the prerequisite.

docs/docs/running-hello-world.md
JohnJoser3: Many readers will skim the intro note and jump to the steps.

Rewrote as: 

To run the client:
1. Open another command line interface.
2. Go to the hello-world directory and run the client.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: I have added a new subheading. 

I wouldn't add underlines or bold here as it will go against our style guide. This is an ideal piece of content to make an architectural diagram, or even an explainer video. Will mark for such an upgrade later.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Removed this line as part of another review comment.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
-[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world). This app contains the `ReverseEnclave`, 
-which reverses a string provided by the client and passes it back.
+This tutorial describes how to compile and run the sample hello world app, which you can find in the
+[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world).
+This app contains the `ReverseEnclave` function, which reverses a string provided by the client and passes it back.
 
-We are first going to run the app in [mock mode](enclave-modes.md),
-which is the simplest (and least realistic) of the enclave modes.
-We will explore running in other modes [later on](#beyond-mock-mode).
+Follow the instructions to run the app in [mock mode](enclave-modes.md). After running it in mock mode, you
```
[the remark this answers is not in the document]
JohnJoser3: Done.

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
-* You need JDK 17 installed in order to build and run the app.
 * This tutorial assumes you've read and understood the [conceptual overview](enclaves.md)
   and [architecture overview](architecture.md).
+* You need JDK 17 or higher installed. You can download it [here](https://www.oracle.com/java/technologies/downloads/).
 
 ## Compile the sample application
 
-Clone and navigate to the hello world sample.
+Like all Conclave apps, this sample app has an [enclave and a client](architecture.md#primary-entities). The enclave
+runs inside a host, which is also a part of the sample app.
```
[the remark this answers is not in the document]
JohnJoser3: Shifted this line along with the link to `Run the host` section.

docs/docs/running-hello-world.md
```diff
@@ -1,55 +1,56 @@
...
 ## Run the host
 
-You can run the host like any other executable JAR with the command
+1. Run the host with the command:
 ```bash
 java -jar host/build/libs/host-mock.jar
 ```
 
-The sample uses Conclave web host. This is a Spring Boot application so you will see the Spring logo as the web server starts up.
+The sample uses the Conclave web host. This is a Spring Boot application. So, you will see the Spring logo when the web
```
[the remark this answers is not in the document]
JohnJoser3: Done with slightly different wording.

docs/docs/running-hello-world.md
```diff
@@ -62,7 +63,8 @@ The sample uses Conclave web host. This is a Spring Boot application so you will
 ```
```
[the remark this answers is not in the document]
JohnJoser3: Thanks, and done.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-available inside the enclave. Our client will use the "Session encryption key" to encrypt data to send into the enclave.
+This output is the [remote attestation](enclaves.md#remote-attestation), an object which proves certain information
+about the enclave. The private key corresponding to the "Session encryption key" (printed above) is only available
+inside the enclave. The client will use the "Session encryption key" to encrypt data to send to the enclave.
+
+After the server starts up, it will be ready to communicate with the client on http://localhost:8080.
 
-Once the server is done starting up it will be ready to communicate with the client on http://localhost:8080.
-You can proceed to [Run the client](#run-the-client) when you see the following output:
+2. Proceed to [Run the client](#run-the-client) when you see the following output:
```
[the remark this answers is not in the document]
JohnJoser3: Done with slightly different wording.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+    The sample uses Conclave Mail which handles encryption and authentication. But the web host protocol uses
+    unencrypted request headers. So, it is important to use HTTPS for anything other than internal development.
+    Setting up an HTTPS connection is beyond the scope of this tutorial. You can try configuring Spring Boot or 
+    setting up a reverse proxy such as Nginx or Apache. See [Conclave web host](conclave-web-host.md) for more 
+    information.
 
 ## Run the client
 
-Run the client with
+After running the host:
```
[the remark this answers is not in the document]
JohnJoser3: Done.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+
+1. Open another command line interface.
+2. Navigate to the hello-world sample.
```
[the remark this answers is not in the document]
JohnJoser3: Done with slightly different wording.

docs/docs/running-hello-world.md
```diff
@@ -73,216 +75,235 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-6. The client sent the encrypted message to the host's REST API.
-7. The host passed the encrypted message to the enclave.
-8. The enclave decrypted the client's message, reversed the string, and encrypted the response using the client's
+
+You have run your first enclave successfully.
+
+The following sequence of events produces the above output:
+
+1. The host loads the enclave, which generates a remote attestation report.
+2. The client connects to the host and downloads the attestation report.
```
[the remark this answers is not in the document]
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Changed.

docs/docs/running-hello-world.md
JohnJoser3: Removed and linked to constraints.md

docs/docs/running-hello-world.md
JohnJoser3: Changed to SIMULATION.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: Done.

docs/docs/running-hello-world.md
JohnJoser3: I removed this section to declutter the tutorial, as part of another comment.

docs/docs/running-hello-world.md
```diff
@@ -2,25 +2,19 @@
...
 
-This tutorial describes how to compile and run the sample hello world app, which you can find in the
-[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world).
-This app contains the `ReverseEnclave` function, which reverses a string provided by the client and passes it back.
-
-Follow the instructions to run the app in [mock mode](enclave-modes.md). After running it in mock mode, you
-can proceed to [run the app in other modes](#beyond-mock-mode).
+This tutorial describes how to compile and run the
+[sample hello world app](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world).
+This app contains an enclave `ReverseEnclave`, which reverses a string provided by the client and passes it
```
shamsasari, line 7: This application contains an enclave which reverses ...

docs/docs/running-hello-world.md
```diff
@@ -2,25 +2,19 @@
...
+back.
 
 If you get stuck at any step, please [talk to us on Discord](https://discord.gg/zpHKkMZ8Sw).
 
 ## Prerequisites
 
-* This tutorial assumes you've read and understood the [conceptual overview](enclaves.md)
-  and [architecture overview](architecture.md).
-* You need JDK 17 or higher installed. You can download it [here](https://www.oracle.com/java/technologies/downloads/).
+* You need JDK 17. You can download JDK 17 [here](https://www.oracle.com/java/technologies/downloads/).
```
shamsasari, line 14: You need JDK 17, which you can download here

docs/docs/running-hello-world.md
```diff
@@ -118,42 +124,25 @@ Reversing `reverse-me` gives `em-esrever`
 
 You have run your first enclave successfully.
 
-The following sequence of events produces the above output:
+
+## Sequence of events
+
+The sample app reversed the string by these steps:
```
shamsasari, line 130: The application securely and confidentially reversed the string by these step

shamsasari (changes requested):
We're using the term "app" instead of "application". Is that good? It feels lazily to me.

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
JohnJoser3: I have removed this warning call-out.

JohnJoser3:
> We're using the term "app" instead of "application". Is that good? It feels lazily to me.

I don't have a strong opinion on this. 'App' sounds more casual and 'application' sounds more formal. I will update it to app. :)

docs/docs/running-hello-world.md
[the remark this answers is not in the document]
JohnJoser3: Defaulted to Linux.

docs/docs/writing-hello-world.md
JohnJoser3: I have changed the link to 1.3.

I feel it's better to have the URL here.

docs/docs/writing-hello-world.md
JohnJoser3: Removed it and added an instruction to generate a signing key along with a link to more details.

docs/docs/writing-hello-world.md
JohnJoser3: Deleted.

docs/docs/writing-hello-world.md
JohnJoser3: Corrected.

docs/docs/writing-hello-world.md
JohnJoser3: This section might fit in Conclave Architecture or a new page for Remote Attestation. 

However, I think we will keep it here for now and lift and shift it when we edit Conclave architecture or create a new page for Remote Attestation.

docs/docs/writing-hello-world.md
JohnJoser3: Done.

docs/docs/writing-hello-world.md
JohnJoser3: Edited.

docs/docs/writing-hello-world.md
shamsasari: Can you update https://r3-cev.atlassian.net/wiki/spaces/CBU/pages/4123394057/GA+Release+Process to mention this  page as well.

docs/docs/running-hello-world.md
```diff
@@ -1,68 +1,80 @@
...
+# Compiling and running your first enclave
 
 ## Introduction
 
-This tutorial describes how to compile and run the sample hello world app, which you can find in the 
-[Conclave tutorials repository](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world). This app contains the `ReverseEnclave`, 
-which reverses a string provided by the client and passes it back.
+This tutorial describes how to compile and run the
+[sample hello world app](https://github.com/R3Conclave/conclave-tutorials/tree/HEAD/hello-world).
+This application contains an enclave `ReverseEnclave`, which reverses a string provided by the client and passes it
```
shamsasari, line 7: Remove "ReverseEnclave". So it would read "This application contains an enclave which reverses ..."

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+   public key.
+9. The enclave passes the encrypted response to the host, which forwards it to the client in an HTTP response.
 
 ## Beyond mock mode
-Eventually you will want to run your application in one of the other [enclave modes](enclave-modes.md), namely simulation,
-debug, or release. This section describes how to build and run the sample in these other modes.
 
-**First, make sure that your system fulfills the
-[system requirements](enclave-modes.md#system-requirements) for your desired mode.**
+In the previous section, you run the sample application in [mock mode](enclave-modes.md).
```
shamsasari, line 141: Can you add, ", which as the name suggests isn't at all secure".

docs/docs/running-hello-world.md
shamsasari: Can we have a quick sentence/addition that says that only release mode is secure and has SGX guarantees. I'm sure this is mention in enclave-modes.md, but if modes are being introduced here, then the most important mode of them all needs to be highlighted. Essentially, we don't want developers thinking that all the modes are created equal.

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-debug, or release. This section describes how to build and run the sample in these other modes.
 
-**First, make sure that your system fulfills the
-[system requirements](enclave-modes.md#system-requirements) for your desired mode.**
+In the previous section, you run the sample application in [mock mode](enclave-modes.md). 
+This section describes how to build and run the sample in other [enclave modes](enclave-modes.md), namely
+simulation, debug, and release.
+
+*First, ensure that your system fulfills the [system requirements](enclave-modes.md#system-requirements) for
+your desired mode.*
```
shamsasari, line 146: Not "desired mode", which sort of implies that debug/simulation are happy compromises to release. We need to quickly, but firmly, state here that only release mode gives you confidential guarantees of SGX.

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+      `docker <command> --help` on the Docker command line interface for more information.
+    3. Run the host and client as you did for mock mode. The only differences are that the host must be run from inside
+      the container, and the host JAR is named `host-simulation.jar` rather than `host-mock.jar`.
+    4. Press `CTRL+D` to exit the container when finished.
+
+    Alternatively, you can also install the Windows subsystem for Linux (2) and download Ubuntu 20.04 from the Windows
+    store. From there, you can launch an Ubuntu 20.04 shell from the start menu and build or run Conclave applications
+    as in a native Linux environment.
+    Instructions for installing WSL-2 can be found [here](https://docs.microsoft.com/en-us/windows/wsl/install).
+    Please note that this method hasn't been extensively tested.
```
shamsasari, line 280: Mention that they should reach out to us if they encounter issues with using WSL, otherwise it doesn't look nice that we're just saying we haven't tested it.

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
     ```text
-    This platform does not support hardware enclaves: SGX_DISABLED_UNSUPPORTED_CPU: SGX is not supported by the CPU in this system
+    This platform does not support hardware enclaves: SGX_DISABLED_UNSUPPORTED_CPU: SGX is not supported by the 
+    CPU in this system
     ```
-    Don't worry, you will still be able to use simulation mode even if you see this message.
+Don't worry, you will still be able to use simulation mode even if you see this message.
 
 ### Run the client in other modes
 The client build is independent of the mode. The only difference is that the enclave will have a different signing key
```
shamsasari, line 294: signing key -> code signer

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-when you are not using mock mode, which will be reflected in the Enclave Constraint provided to the client.
+when you are not using mock mode, which will be reflected in the enclave constraint provided to the client.
```
shamsasari, line 295: It's not just mock mde necessarily. For example, we talk about release mode have an external code signing key. That will mean it will be different from debug and simulation. So I would rephrase his sentence to something like

"The only difference is that the enclave might have a different code coder, ..."

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
-  - Product ID: 1
-  - Revocation level: 0
-```
 
-If you haven't built the client already, build it the same way as you would for mock mode
 
 
+=== "Linux/macOS"
+    
+    1. When you run the host, you should see the new signing key in the output:
```
shamsasari, line 303: coder signer

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+    - Session signing key: 302A300506032B6570032100D23DD5C05A37CB5B6ED50EA1501E55ABF0EF85B50A97A69D0C3F4F84372AF928
+    - Session encryption key: 42CF5E2457B19A9E4FA3716F40CDF6B07A3EEC95D1AFE29C6F1DE99FD0DC647C
+    - Product ID: 1
+    - Revocation level: 0
+    ``` 
+    2. If you haven't built the client already, build it the same way as in mock mode:
+    ```bash
+    ./gradlew :client:shadowJar
+    ```
+    3. Run the client using the signing key from the host's attestation report
```
shamsasari, line 318: code signer

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+    ```bash
+    java -jar client/build/libs/client.jar "S:4924CA3A9C8241A3C0AA1A24A407AA86401D2B79FA9FF84932DA798A942166D4 PROD:1 SEC:INSECURE" reverse-me
+    ```
 === "Windows"
+    
+    1. When you run the host, you should see the new signing key in the output:
```
shamsasari, line 324: ditto

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
+    - Session encryption key: 42CF5E2457B19A9E4FA3716F40CDF6B07A3EEC95D1AFE29C6F1DE99FD0DC647C
+    - Product ID: 1
+    - Revocation level: 0
+    ``` 
+    2. If you haven't built the client already, build it the same way as in mock mode:
     ```bash
     gradlew.bat :client:shadowJar
     ```
-=== "macOS / Linux"
+    3. Run the client using the signing key from the host's attestation report
```
shamsasari, line 339: ditto

docs/docs/running-hello-world.md
```diff
@@ -73,265 +85,289 @@ The web server loads the enclave and waits for requests from the client. Amongst
...
 
-[:fontawesome-brands-discord: Join our discord server](https://discord.com/invite/dDaBGqGPre){: .md-button }
+[:fontawesome-brands-discord: Join our discord server](https://discord.gg/zpHKkMZ8Sw){: .md-button }
 
 [:fontawesome-solid-paper-plane: Join conclave-discuss@groups.io](https://groups.io/g/conclave-discuss){: .md-button }
 
-[:fontawesome-solid-paper-plane: Email us directly](mailto:conclave@r3.com){: .md-button }
+[:fontawesome-solid-paper-plane: Email us](mailto:conclave@r3.com){: .md-button }
 
 ## Appendix: Summary of Docker command options
```
shamsasari, line 363: We should not be explaining docker command options! These are already well documented on Docker's own docsite. Let's please remove this.

docs/docs/writing-hello-world.md
```diff
@@ -1,45 +1,47 @@
...
-* Modifying the files to create an implementation of the Conclave *sample app* which we explored in the
+* Set up a new Conclave project using [Conclave Init](conclave-init.md).
+* Modify the files to create an implementation of the Conclave *sample app* which we explored in the
   [previous tutorial](running-hello-world.md).
 
-We will be leveraging Conclave's built-in web host to write our Conclave app. To learn more about this, see
-[Conclave web host](conclave-web-host.md). You may want to implement your own host if you need custom functionality or
-you want to incorporate it into an existing server application of some kind. To learn more about that, see
-[Writing your own enclave host](writing-your-own-enclave-host.md).
+This tutorial uses [Conclave web host](conclave-web-host.md) to write a Conclave app.
```
shamsasari, line 14: Do we even need this sentence? the web host is already mentioned when further down where it's relevant. No need to talk about something the user will have no idea about.

docs/docs/writing-hello-world.md
```diff
@@ -223,37 +185,36 @@ Assessed security level at 2020-07-17T16:31:51.894697Z is INSECURE
...
+enclave. The clients can audit the enclave by comparing the value they get in the
+[`EnclaveInstanceInfo`](api/-conclave%20-core/com.r3.conclave.common/-enclave-instance-info/index.html) against what 
+the build process prints out.
+
+!!!Note
+
+    1. You can get all this data using individual getters on the
+       [`EnclaveInstanceInfo`](api/-conclave%20-core/com.r3.conclave.common/-enclave-instance-info/index.html). So 
+       you don't need to parse the output of `toString`.
+    2. You can easily build mock attestations in your tests using `EnclaveInstanceInfo` interface.
```
shamsasari, line 200: Delete this bullet point. `EnclaveInstanceInfo` isn't designed to be extended by the user.

docs/docs/writing-hello-world.md
```diff
@@ -223,37 +185,36 @@ Assessed security level at 2020-07-17T16:31:51.894697Z is INSECURE
...
+    3. When not in simulation mode, the timestamp is signed by Intel.
```
shamsasari, line 201: Also when not in mock mode, i.e. it's signed by Intel for debug and release modes. However this is general point not just limited to the timestamp. Only in debug and release modes is the remote attestation signed by Intel. For mock and simulation modes the attestation is a fake mock object. Perhaps this general point should be explained in a normal paragraph and not in a "note" box or bullet point.

docs/docs/writing-hello-world.md
```diff
@@ -423,7 +366,7 @@ EnclaveHost mockHost = EnclaveHost.load("com.r3.conclave.sample.enclave.ReverseE
 mockHost.start(null, null);
 ```
 
-Conclave will detect that the enclave class is on the classpath and will start the enclave in mock mode. You
+Conclave detects the enclave class on the classpath and starts the enclave in mock mode. You
 can obtain the enclave instance using the
 [`EnclaveHost.mockEnclave`](api/-conclave%20-core/com.r3.conclave.host/-enclave-host/get-mock-enclave.html) property.
```
shamsasari, line 372: Would it make sense to point the developer to a sample mock unit test?

https://github.com/R3Conclave/conclave-tutorials/blob/HEAD/hello-world/enclave/src/test/java/com/r3/conclave/sample/enclave/MockTest.java

docs/docs/writing-hello-world.md
```diff
@@ -464,42 +406,30 @@ public class NativeTest {
...
 ```
 
-Or you can use a mock enclave and test on a non-Linux platform by removing `@EnabledOnOs(OS.LINUX)` and by running this
-command:
+To use a mock enclave and test on a non-Linux platform, remove `@EnabledOnOs(OS.LINUX)` and run:
 
-```
+```bash
 ./gradlew -PenclaveMode=mock host:test
 ```
```
shamsasari, line 430: Likewise for the native tests: https://github.com/R3Conclave/conclave-tutorials/blob/HEAD/hello-world/host/src/test/java/com/r3/conclave/sample/host/NativeTest.java

shamsasari (commented):
Just finished my second pass. After this I think I'm good and happy with the changes.

docs/docs/writing-hello-world.md
```diff
@@ -262,20 +222,7 @@ java -jar host/build/libs/host-mock.jar
...
-
-Actually, the first two steps are done for you when using an
-[`EnclaveClient`](api/-conclave%20-core/com.r3.conclave.client/-enclave-client/index.html) object. `EnclaveClient` handles the encryption and
-decryption of Mail for you and provides a simple interface for sending and receiving Mail. However, it doesn't know
-_how_ to transport the mail, which is where [`EnclaveTransport`](api/-conclave%20-core/com.r3.conclave.client/-enclave-transport/index.html)
-comes in. Since we're connecting to the host web server, we'll be using
-[`WebEnclaveTransport`](api/-conclave%20-core/com.r3.conclave.client.web/-web-enclave-transport/index.html) as our transport.
-
-Replace the client code at `client/src/.../ReverseEnclaveClient.java`
+Replace the client code at `client/src/.../ReverseEnclaveClient.java` with:
```
bonifazi-r3, line 225: Maybe it is better to provide the complete path? Not sure if there is a convention

docs/docs/writing-hello-world.md
```diff
@@ -223,37 +183,37 @@ Assessed security level at 2020-07-17T16:31:51.894697Z is INSECURE
...
+
+
+An instance has a [security assessment](api/-conclave%20-core/com.r3.conclave.common/-enclave-security-info/index.html),
+which changes when infrastructure vulnerabilities are discovered (even without any changes in the host or enclave). 
+This sample enclave isn't secure yet because it's running in simulation mode. An enclave can be `SECURE`, `STALE`, 
+or `INSECURE`. An assessment of `STALE` means a software/firmware/microcode update that improves security is 
+available. The client needs to define a time span by which the remote enclave operator must upgrade when the 
+security assessment turns `STALE`.
+
+An attestation doesn't expire. As the SGX ecosystem is constantly changes, the client code needs to define a frequency
```
bonifazi-r3, line 207: `As the SGX ecosystem is constantly changes`, changes=changed?

shamsasari (approved):
Thanks for this John.

docs/docs/writing-hello-world.md
```diff
@@ -223,37 +183,37 @@ Assessed security level at 2020-07-17T16:31:51.894697Z is INSECURE
...
+which changes when infrastructure vulnerabilities are discovered (even without any changes in the host or enclave). 
+This sample enclave isn't secure yet because it's running in simulation mode. An enclave can be `SECURE`, `STALE`, 
+or `INSECURE`. An assessment of `STALE` means a software/firmware/microcode update that improves security is 
+available. The client needs to define a time span by which the remote enclave operator must upgrade when the 
+security assessment turns `STALE`.
+
+An attestation doesn't expire. As the SGX ecosystem is constantly changes, the client code needs to define a frequency 
+with which it expects the host code to refresh the
+[`EnclaveInstanceInfo`](api/-conclave%20-core/com.r3.conclave.common/-enclave-instance-info/index.html). Currently, 
+the host code is refreshed by restarting the enclave.
```
bonifazi-r3, line 210: Minor note: as we are talking about frequency in the sentence before, maybe it is better a `is refreshed every time the enclave restarts` or similar?

bonifazi-r3 (changes requested):
Looks very good! There just a typo, I believe, and minor comments

docs/docs/writing-hello-world.md
JohnJoser3: We use a similar path in another place in the doc too. I think it's okay. It also allows flexibility if any subfolder changes in the future.

docs/docs/writing-hello-world.md
JohnJoser3: Oh! Changed to `As the SGX ecosystem constantly changes`

docs/docs/writing-hello-world.md
JohnJoser3: Changed to `the host code refreshes every time the enclave restarts`.

bonifazi-r3 (approved):
Great work!

PR #22, opened 2022-08-31
CON-1138: Make links readable on the docs site

JohnJoser3:
Currently, links are not distinguishable from standard text on the docs site. Only when a user hovers over the link text the user knows it's a link.

I have changed the color of links by adding a CSS variable '--md-typeset-a-color' in the r3-branding.css file.  I have used a lighter version of teal, which wouldn't stand out too much while being readable. The hex color code is #006c94. Needs to test how it looks when published.

### DIFF
--- a/docs/docs/css/r3-branding.css
+++ b/docs/docs/css/r3-branding.css
@@ -13,6 +13,7 @@ See:
   --md-primary-fg-color--light:        #53585F;
   --md-primary-fg-color--dark:         #00161E;
   --md-accent-fg-color:                #EC1D24;
+  --md-typeset-a-color:                #006c94;
 }
 
 :root > * {
--- a/docs/docs/css/r3-branding.css
+++ b/docs/docs/css/r3-branding.css
@@ -13,7 +13,7 @@ See:
   --md-primary-fg-color--light:        #53585F;
   --md-primary-fg-color--dark:         #00161E;
   --md-accent-fg-color:                #EC1D24;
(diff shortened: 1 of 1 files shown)


shamsasari:
The colour difference is almost non-existent for me, but that's probably because I'm colour blind: https://r3-conclave.artifacts.teamcity.com/presignedTokenAuth/d578c98b-474e-47e9-8d0b-c578ca8c7e1d/repository/download/ConclaveCore_SdkBuild/4097:id/docs.zip!/running-hello-world.html

shamsasari:
I've also added @danibentrup in case she wants to make sure what we choose is inline with our branding.

JohnJoser3:
> The colour difference is almost non-existent for me, but that's probably because I'm colour blind: https://r3-conclave.artifacts.teamcity.com/presignedTokenAuth/d578c98b-474e-47e9-8d0b-c578ca8c7e1d/repository/download/ConclaveCore_SdkBuild/4097:id/docs.zip!/running-hello-world.html

danibentrup:
> I've also added @danibentrup in case she wants to make sure what we choose is inline with our branding.

@shamsasari I don't have access to TeamCity. Should I get an account to preview this?

JohnJoser3:
Now the links have a readable color closer to our brand colors and distinguishable from black.

shamsasari:
@danibentrup, just raise an IT ticket to get access as you can only login via SSO.

shamsasari (approved):
Thanks John, looks good.

PR #32, opened 2022-09-06
Changes to azure.md and DC4s_v3 screenshot.

JohnJoser3:
Edits to correct azure.md doc


docs/docs/azure.md
```diff
@@ -41,28 +38,23 @@ question mark is the size. Other distributions should work as long as they are o
...
+1. Check that the `enclave` device is present in the `/dev/sgx/` directory.
+2. Add your user into `sgx_prv` group to give it access to SGX.
 
 ```sh
 sudo usermod -aG sgx_prv $USER
 ```
-5. Log out from the VM and log in again.
+3. Log out from the VM and log in again.
 
 You have set up an Azure VM with the latest attestation protocols.
```
bonifazi-r3, line 54: @shamsasari @anien  While reading this documents, I did not understand why we mention attestation in the first part but then we do not actually install it and we tell the user to "skip ... if you are only looking for a quick setup"

I do not have the full context of this page, but at this specific line of these instructions we do not have attestation available, wouldn't be better to say something like
```
You have set up an Azure VM with SGX.
If you also want to setup attestation, please follow the instructions below.

Comment (anien) [reply docs/docs/azure.md:54]: Agree. When we have a new user coming to quickly setup Azure VM and run for example hello world sample, we should keep this short. Especially as we advice to use DC4s_v3 images that are ready to go.

File: docs/docs/azure.md [hunk 2]
```diff
@@ -41,28 +38,23 @@ question mark is the size. Other distributions should work as long as they are o
...
 
-You can skip the rest of this document if you are only looking for a quick set up.
```
anien, line 63: Why removing this line?
I think we need to tell user they don't need to continue if they are looking for a quick set up.

docs/docs/azure.md
```diff
@@ -41,28 +38,23 @@ question mark is the size. Other distributions should work as long as they are o
...
 
-## Background: The DCAP protocol
+## The DCAP protocol
```
shamsasari, line 57: This section is no longer relevant since Azure has everything setup. Can we move this to a separate page for now.

The one piece of information developers will need to know when setting up an SGX VM on Azure. and which should be mentioned in this page, is that they will need to use DCAP for attestation. So that means from a code point of view, they will need to use the `AttestationParameters.DCAP()` class when starting the enclave via `EnclaveHost.start`.

shamsasari (changes requested):

docs/docs/azure.md
JohnJoser3: Moved the section to a new page 'dcap.md'. Added a short instruction to use use the `AttestationParameters.DCAP()` class when starting the enclave via `EnclaveHost.start` for DCAP.

docs/docs/azure.md
```diff
@@ -41,28 +38,23 @@ question mark is the size. Other distributions should work as long as they are o
...
 
 After you have logged on to the VM:
 
-1. Check that the `enclave` device is present in the `/dev/sgx/` or `/dev/sgx_*` directory
-2. Check driver version `dmesg | grep sgx`. Conclave requires driver version 1.33+
-3. If either check fails:
-    * Download the [driver](https://download.01.org/intel-sgx/latest/dcap-latest/linux/distro/ubuntu20.04-server/)
-    * Follow the [install instructions](https://download.01.org/intel-sgx/sgx-dcap/1.8/linux/docs/Intel_SGX_DCAP_Linux_SW_Installation_Guide.pdf)
-4. Add your user into `sgx_prv` group to give it access to SGX.
+1. Check that the `enclave` device is present in the `/dev/sgx/` directory.
```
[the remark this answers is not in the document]
JohnJoser3: Added back /dev/sgx_*

docs/docs/azure.md
```diff
@@ -4,9 +4,6 @@
     Azure does not guarantee access to the same machine on reboot. On reboot, you might lose encrypted secrets for a particular enclave.
     To prevent loss of data, consider using the [Key Derivation Service](kds-detail.html) (KDS).
```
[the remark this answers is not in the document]
JohnJoser3: Rephrased as "To prevent data loss, you need to use the..."

docs/docs/azure.md
```diff
@@ -41,28 +38,23 @@ question mark is the size. Other distributions should work as long as they are o
     * Pick a size that's got plenty of RAM. For example, you might want to click "Change size" to find `DC4s_v3` type
     * Ensure that the public inbound ports are open for SSH access
 
-    ![](images/create-d4sv3-vm.png)
+    ![](images/DC4S_V3 Screenshot.png)
```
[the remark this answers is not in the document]
JohnJoser3: Changed.

docs/docs/azure.md
JohnJoser3: Moved the entire DCAP section to another page.

docs/docs/azure.md
JohnJoser3: Moved the entire DCAP section to another page.

docs/docs/azure.md
```diff
@@ -38,132 +38,21 @@ question mark is the size. Other distributions should work as long as they are o
...
-
-#### Azure client plugin
+You have set up an Azure VM with SGX.
 
-* Identify the currently installed DCAP client plugin. It will always have one of the following names: `libdcap_quoteprov.so.1` or `libdcap_quoteprov.so`. You might find other similarly named files, but they won't be used as a plugin.
-```sh
-ls /usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so*
-ls /usr/lib/libdcap_quoteprov.so*
-```
+You need to use the DCAP protocol for attestation. For this, you need to use the `AttestationParameters.DCAP()` class
```
shamsasari, line 55: Can you reference the class in the API docs.

docs/docs/azure.md
```diff
@@ -38,132 +38,21 @@ question mark is the size. Other distributions should work as long as they are o
...
-#### Azure client plugin
+You have set up an Azure VM with SGX.
 
-* Identify the currently installed DCAP client plugin. It will always have one of the following names: `libdcap_quoteprov.so.1` or `libdcap_quoteprov.so`. You might find other similarly named files, but they won't be used as a plugin.
-```sh
-ls /usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so*
-ls /usr/lib/libdcap_quoteprov.so*
-```
+You need to use the DCAP protocol for attestation. For this, you need to use the `AttestationParameters.DCAP()` class 
+when starting the enclave via `EnclaveHost.start`.
```
shamsasari, line 56: likewise

docs/docs/dcap.md
```diff
@@ -0,0 +1,111 @@
+## The DCAP protocol
+
+There are two protocols for establishing what code runs in an enclave: EPID and DCAP. EPID is an older
+protocol designed for consumer applications and includes some sophisticated privacy features. For servers
+where the IP address doesn't need to be hidden (because it's public in DNS, to begin with), these features
+aren't helpful, and thus there is DCAP (_Data Center Attestation Primitives_). DCAP requires more modern hardware
+but is otherwise simpler and more robust. You may also see DCAP referred to as "ECDSA attestation".
+
+!!! important
+Intel does not provide EPID attestation support on any Xeon Scalable processors including the Ice Lake Xeon CPUs and any future
```
shamsasari, line 10: This is not rendered correctly

docs/docs/azure.md
```diff
@@ -50,9 +54,9 @@ sudo usermod -aG sgx_prv $USER
...
 
-You have set up an Azure VM with SGX.
-
-You need to use the DCAP protocol for attestation. For this, you need to use the `AttestationParameters.DCAP()` class 
-when starting the enclave via `EnclaveHost.start`.
-
+You have set up an Azure VM with the latest attestation protocols.
 
+You need to use the DCAP protocol for attestation. For this, you need to use the
+[`AttestationParameters.DCAP()`](https://docs.conclave.net/api/-conclave%20-core/com.r3.conclave.host/-attestation-parameters/-d-c-a-p/-attestation-parameters.-d-c-a-p.html) class when starting the
```
shamsasari, line 60: Remove the `()`, sorry

docs/docs/azure.md
JohnJoser3: Done.

docs/docs/azure.md
JohnJoser3: Done.

docs/docs/dcap.md
JohnJoser3: Done. Also corrected other call-out boxes.

docs/docs/azure.md
JohnJoser3: Done.

shamsasari (changes requested):
Can we make the azure page more visible in the "Operations" section. you can only get to it by clicking on the link in the first bullet point. It should either be in list on the left, or we should merge it with the machine setup page. 

Deploying to Azure is going to be the most common way to deploy a Conclave app and we don't make it super easy and obvious to find this information.

JohnJoser3:
> 

I have removed the unnecessary branching.

Now, the Cloud Deployment (Deploy to Azure) page can be seen without clicking on the parent folder.

docs/mkdocs.yml
```diff
@@ -43,10 +44,9 @@ nav:
   - Release notes: release-notes.md
   - Operations:
     - Server deployment:
-      - Machine setup:
-         - Overview: machine-setup.md
-         - Cloud Deployment: azure.md
-         - Self-Hosting: ias.md
+        - Machine Setup: machine-setup.md
+        - Cloud Deployment: azure.md
```
shamsasari, line 48: Can we have this the first in the line as it's going to be the most common way to deploy.

JohnJoser3: The machine-setup page provides an overview of azure deployment and IAS. So, if we change the order, it would affect the flow. I recommend to keep the order as is.

shamsasari: Whilst IAS isn't deprecated by Intel, it is no longer the default remote attestation protocol, which is now DCAP, and that is what's available on Azure. So the problem with the current page layout is that IAS is given Operations front page whilst the more common and default Azure/DCAP has to accessed via another click.

We definitely need to fix this, but I'm happy for you to restructure as part as of your general rewrite of the docsite. I appreciate the original PR was raised to fix the screenshots and azure docs. I'll approve to get this merged.

shamsasari (approved):

docs/mkdocs.yml
JohnJoser3: I will add a ticket to the backlog to fix this.

PR #42, opened 2022-09-26
CON-1161 - Integrated gramine Conclave and refactored the code

filipesoliveira:
It is now possible to run Conclave using Gramine but without sgx. This is only experimental and more changes will be added in future commits. To use gramine, Conclave must be built in mock mode and the environment variable CONCLAVE_GRAMINE_ENABLE must be set.

The class EnclaveScanner was created to group together all the functions responsible for searching for enclaves.

### DIFF
--- a/conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
+++ b/conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
@@ -424,10 +424,10 @@ class EnclaveHost private constructor(
         if (hostStateManager.state is Started) return
         hostStateManager.checkStateIsNot<Closed> { "The host has been closed." }
 
-        // For now, there is no enclave for gramine instead the mock enclave is initialised.
+        // For now, there is no enclave for Gramine instead the mock enclave is initialised.
         // This is OK because no VM is started in Mock mode and this will allow us to integrate Gramine with Conclave
         // iteratively without causing issues to the normal operation of Conclave.
-        // Start gramine only if the environment variable is set
+        // Start Gramine only if the environment variable is set
         if(Gramine.isGramineEnabled())
         {
             if(enclaveHandle !is MockEnclaveHandle) {
--- a/conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
+++ b/conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
@@ -126,7 +126,7 @@ class EnclaveHost private constructor(
         @JvmStatic
         @Throws(EnclaveLoadException::class, PlatformSupportException::class)
         fun load(enclaveClassName: String, mockConfiguration: MockConfiguration?): EnclaveHost {
-            return createEnclaveHost(findEnclave(enclaveClassName), mockConfiguration)
+            return createEnclaveHost(EnclaveScanner.findEnclave(enclaveClassName), mockConfiguration)
         }
 
         /**
@@ -146,7 +146,7 @@ class EnclaveHost private constructor(
         @JvmStatic
         @Throws(EnclaveLoadException::class, PlatformSupportException::class)
         fun load(mockConfiguration: MockConfiguration?): EnclaveHost {
-            return createEnclaveHost(findEnclave(), mockConfiguration)
+            return createEnclaveHost(EnclaveScanner.findEnclave(), mockConfiguration)
         }
 
         /**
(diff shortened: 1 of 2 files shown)


conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -537,6 +424,18 @@ class EnclaveHost private constructor(
         if (hostStateManager.state is Started) return
         hostStateManager.checkStateIsNot<Closed> { "The host has been closed." }
 
+        // For now, there is no enclave for gramine instead the mock enclave is initialised.
+        // This is OK because no VM is started in Mock mode and this will allow us to integrate Gramine with Conclave
+        // iteratively without causing issues to the normal operation of Conclave.
+        // Start gramine only if the environment variable is set
```
anien, line 430: Just a minor - keep Gramine with uppercase 'G' consistent.

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -537,6 +424,18 @@ class EnclaveHost private constructor(
...
+        if(Gramine.isGramineEnabled())
+        {
+            if(enclaveHandle !is MockEnclaveHandle) {
+                throw Exception("Gramine cannot be started in non-mock modes")
+            }
+            Gramine.start()
+        }
+
```
anien, line 438: If you have environmental variable set and you use non-mock mode, will it fail? Shouldn't it just ignore it in that case and maybe write a warning instead? Just an idea, not really fussy about the behaviour at this stage.

filipesoliveira: I thought it would be better to have an explicit error for now to be easier to understand what is happening. I am happy to change the code if you prefer the log message.

anien: Let's leave it for now.

conclave-host/src/main/kotlin/com/r3/conclave/host/Gramine.kt
```diff
@@ -0,0 +1,116 @@
...
+        val envValue = System.getenv("CONCLAVE_GRAMINE_ENABLE") ?: return false
+        return envValue.uppercase() == "TRUE"
+    }
+
+    private fun generateManifestFile(destination: Path) {
+        val file = File(destination.absolutePathString() + "/" + "bash.manifest")
+        writeToFile(manifestContent.toByteArray(), file)
+    }
+
+    private fun writeToFile(content: ByteArray, file: File) {
```
anien, line 48: I think we should try to use Path API where possible instead of java.io.File

filipesoliveira: What would the issue be if File is used? The file will not exist if an object of type Path is used.

anien: I meant in general in this file. I believe we are just trying to use new Path API in general, instead of File API.

anien (approved):
LGTM now.

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -537,6 +424,18 @@ class EnclaveHost private constructor(
         if (hostStateManager.state is Started) return
         hostStateManager.checkStateIsNot<Closed> { "The host has been closed." }
 
+        // For now, there is no enclave for Gramine instead the mock enclave is initialised.
+        // This is OK because no VM is started in Mock mode and this will allow us to integrate Gramine with Conclave
+        // iteratively without causing issues to the normal operation of Conclave.
+        // Start Gramine only if the environment variable is set
+        if(Gramine.isGramineEnabled())
+        {
+            if(enclaveHandle !is MockEnclaveHandle) {
```
shamsasari, line 433: Please follow the Kotling coding style.

conclave-host/src/main/kotlin/com/r3/conclave/host/Gramine.kt
```diff
@@ -0,0 +1,109 @@
+package com.r3.conclave.host
```
shamsasari, line 1: This package is part of the public API namespace, and this being an internal class cannot reside here. Move this to the internal package.

PR #45, opened 2022-09-30
[CON-557] INTEL-SA-00398 - Intel® SGX DCAP Software Advisory

ac101m:
# Changes
- Update DCAP plugin to version 1.11.2


ac101m:
We were technically already above the required version for this SA (1.10 vs 1.6), but I decided to bump the plugin version anyway. Should also get rid of that annoying logging "error" that we see in the terminal sometimes.

filipesoliveira (approved):

PR #46, opened 2022-10-03
CON-806 - Review INSECURE, STALE, and SECURE enclave constraint options

filipesoliveira:
Updated how the security level is assess. If the security level is set to INSECURE then a client can only communicate with an enclave that is running as INSECURE. If the security level is set to STALE then the client can communicate with a STALE or SECURE enclave. If the security level is set to SECURE then the client can only communicate with a SECURE enclave. This prevents clients from connecting to enclaves running in production by mistake or by a malicious host. Check CON-806 for more details.

Updated the unit-tests file.

### DIFF
--- a/conclave-common/src/main/kotlin/com/r3/conclave/common/EnclaveConstraint.kt
+++ b/conclave-common/src/main/kotlin/com/r3/conclave/common/EnclaveConstraint.kt
@@ -145,7 +145,7 @@ class EnclaveConstraint {
     }
 
     private fun checkSecurityLevel(enclave: EnclaveInstanceInfo) {
-        // If the security level required by the client is insecure than the enclave security should be insecure. This prevents
+        // If the security level required by the client is insecure then the enclave security must be insecure. This prevents
         // clients from connecting to enclaves running in production by mistake. This also stops a malicious host from connecting
         // a client to a production enclave in order to corrupt data. Check CON-806 for more details
         if (minSecurityLevel == EnclaveSecurityInfo.Summary.INSECURE) {
--- a/conclave-common/src/main/kotlin/com/r3/conclave/common/EnclaveConstraint.kt
+++ b/conclave-common/src/main/kotlin/com/r3/conclave/common/EnclaveConstraint.kt
@@ -146,7 +146,8 @@ class EnclaveConstraint {
 
     private fun checkSecurityLevel(enclave: EnclaveInstanceInfo) {
         // If the security level required by the client is insecure than the enclave security should be insecure. This prevents
-        // clients from connecting to enclaves running in production by mistake.
+        // clients from connecting to enclaves running in production by mistake. This also stops a malicious host from connecting
+        // a client to a production enclave in order to corrupt data. Check CON-806 for more details
         if (minSecurityLevel == EnclaveSecurityInfo.Summary.INSECURE) {
             checkEnclave(enclave.securityInfo.summary == minSecurityLevel) {
                 "Enclave has a security level of ${enclave.securityInfo.summary} which does not match the required level of $minSecurityLevel."
--- a/conclave-common/src/main/kotlin/com/r3/conclave/common/EnclaveConstraint.kt
+++ b/conclave-common/src/main/kotlin/com/r3/conclave/common/EnclaveConstraint.kt
@@ -134,14 +134,26 @@ class EnclaveConstraint {
             }
         }
 
-        checkEnclave(enclave.securityInfo.summary >= minSecurityLevel) {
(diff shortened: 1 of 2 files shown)


conclave-common/src/main/kotlin/com/r3/conclave/common/EnclaveConstraint.kt
```diff
@@ -146,6 +144,21 @@ class EnclaveConstraint {
         }
     }
 
+    private fun checkSecurityLevel(enclave: EnclaveInstanceInfo) {
+        // If the security level required by the client is insecure than the enclave security should be insecure. This prevents
```
anien, line 148: I would highlight that enclave security lever _must_ be insecure 
_enclave security should -> enclave security level must_

conclave-common/src/test/kotlin/com/r3/conclave/common/EnclaveConstraintTest.kt
```diff
@@ -278,6 +280,54 @@ class EnclaveConstraintTest {
...
+            val constraint = EnclaveConstraint.parse("SEC:$securityLevelRequiredByClient C:${secureEnclave.enclaveInfo.codeHash}")
+            assertDoesNotThrow { constraint.check(secureEnclave) }
+        }
+
+        assertThrows<InvalidEnclaveException> {
+            val securityLevelRequiredByClient = INSECURE
+            val constraint = EnclaveConstraint.parse("SEC:$securityLevelRequiredByClient C:${secureEnclave.enclaveInfo.codeHash}")
+            constraint.check(secureEnclave)
+        }
+    }
```
anien, line 297: Shouldn't parse throw an exception if securityLevelRequiredByClient is STALE?
My understanding is that enclave security level = SECURE => if securityLevelRequiredByClient in {INSECURE, STALE} throws an exception. 
It shouldn't throw an exception only if securityLevelRequiredByClient = SECURE

conclave-common/src/test/kotlin/com/r3/conclave/common/EnclaveConstraintTest.kt
```diff
@@ -278,6 +280,54 @@ class EnclaveConstraintTest {
...
+            val securityLevelRequiredByClient = STALE
+            val constraint = EnclaveConstraint.parse("SEC:$securityLevelRequiredByClient C:${secureEnclave.enclaveInfo.codeHash}")
+            constraint.check(secureEnclave)
+        }
+
+        listOf(INSECURE, SECURE).forEach { securityLevelRequiredByClient ->
+            val constraint = EnclaveConstraint.parse("SEC:$securityLevelRequiredByClient C:${secureEnclave.enclaveInfo.codeHash}")
+            assertThrows<InvalidEnclaveException> { constraint.check(secureEnclave) }
+        }
+    }
```
anien, line 313: Same here, in case of enclave's security level = STALE, then STALE and SECURE shouldn't throw an exception. Only INSECURE should throw an exception.

anien (changes requested):
Some comments in the tests. Not sure why integration tests passed actually.

conclave-common/src/test/kotlin/com/r3/conclave/common/EnclaveConstraintTest.kt
filipesoliveira: I am not sure if I am understanding your comment. If an enclave is classified as secure then a client that sets the constraint as `STALE` or `SECURE` should be able to connect to that enclave. The client sets the minimum security that it expects the enclave to have. However, if the client wants the min. security level to be insecure then the enclave must be classified as insecure.

anien: Nope, that’s not correct. It goes against your PR comment which is correct. 
“If the security level is set to STALE then the client can communicate with a STALE or SECURE enclave. If the security level is set to SECURE then the client can only communicate with a SECURE enclave. “

The security levels are as follows INSECURE <= STALE <= SECURE

filipesoliveira: _“If the security level is set to STALE then the client can communicate with a STALE or SECURE enclave. If the security level is set to SECURE then the client can only communicate with a SECURE enclave. “_
This is what the unit-tests are doing.

anien: Oh sorry, I have mixed and matched the security level of an enclave and client's required security level! Looks good now!

anien (approved):

PR #47, opened 2022-10-03
CON-1127 Upgrade to latest Kotlin 1.7.20

filipesoliveira:


anien (approved):
Looks good.

filipesoliveira:
**Note:**
image

filipesoliveira:
SDK Build (Conclave Core):
- It fails when Kotlin is upgraded to 6.20 or 7.20
- It fails only on TC.
- ./gradlew test works fine on a local machine or an Azure machine
-

shamsasari:
The error is occuring because `com.r3.conclave.enclave.Enclave$Companion.getMailDecryptingStream(Enclave.kt:73)` is trying to invoke the c'tor

`com.r3.conclave.mail.internal.MailDecryptingStream.<init>(Ljava/io/InputStream;Ljava/security/PrivateKey;ILkotlin/jvm/internal/DefaultConstructorMarker;)V`.

It shouldn't be doing this, because this c;tor doesn't exist. The c'tor that does exist is 

`(Ljava/io/InputStream;Ljava/security/PrivateKey;ILcom/r3/conclave/shaded/kotlin/jvm/internal/DefaultConstructorMarker;)V`

i.e. the difference is in the last parameter. The `DefaultConstructorMarker` should be our shaded one (`com.r3.conclave.shaded...`) and not the standard kotlin one.

The question is why isn't the shading process updating `Enclave$Companion.getMailDecryptingStream` to use our shaded c;tor.

shamsasari:
Can you try `./gradlew -i conclave-enclave:shadowJar` and see if the shadow plugin prints out any debug logging. What we want to find out is why it is not rewriting the bytecode for `Enclave.Companion.getMailDecryptingStream` to use the shaded c;tor.

filipesoliveira:
The integration tests are successful when using the Kotlin version 6.20.

filipesoliveira:
Todo:
- Remove the generated enclave_jni.
- Fix the task clean to also delete the enclave_jni that is generated by the cmake.
- Update the dependensOn so the enclave_jni is created before it is needed.
- Undo the changes that are not required.
- Leave a comment about the fix done.

shamsasari:
Huh, merged without approval! I guess it must have been Anna's previous approval?

cpp/build.gradle
```diff
@@ -1,5 +1,11 @@
 apply plugin: "base"
 
+task cleaGeneratedFiles {
```
shamsasari, line 3: typo: clean

build.gradle
```diff
@@ -287,26 +287,36 @@ dokkaHtml.shouldRunAfter mkdocs
...
 for (publishedProject in publishedProjects) {
-    def project = project(publishedProject.projectPath)
+    def project1 = project(publishedProject.projectPath)
 
-    project.pluginManager.apply('java')
-    project.pluginManager.apply('maven-publish')
-    project.pluginManager.apply('signing')
+    project1.pluginManager.apply('java')
+    project1.pluginManager.apply('maven-publish')
+    project1.pluginManager.apply('signing')
```
shamsasari, line 294: What's the reason for this rename?

filipesoliveira: There was a name collision between the function `project`, in line 290 and the variable `project`. This causes in line 307 where the function `project` is required again.

filipesoliveira:
PR https://github.com/R3Conclave/conclave-core-sdk/pull/69 addresses the suggestions made in this PR after it was merged.

build.gradle
shamsasari: I presumed it was due to groovy. Another reason to switch to the kotlin DSL.

PR #57, opened 2022-10-10
[CON-1172] Implement stream call interface for Gramine integration

ac101m:
# Changes
- Implement socket call interface classes and associated tests.

# Notes
The new interfaces are reliable but slow, even over localhost. We will need to investigate named pipes, datagram sockets or memory mapped files to get performance to an acceptable level. This remains as future work.


conclave-common/src/main/kotlin/com/r3/conclave/common/internal/StreamCallInterfaceMessage.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun size() = headerSize + payloadSize
+
+    companion object {
+        fun fromBytes(bytes: ByteArray): StreamCallInterfaceMessage {
+            val buffer = ByteBuffer.wrap(bytes)
+
+            val hostThreadID = buffer.long
+            val callTypeID = buffer.get()
+            val messageTypeID = buffer.get()
+            val messageBytes = buffer.getNullable { getRemainingBytes() }
```
shamsasari, line 28: Just conscious here that we might end up doing extra copying, all to wrap the payload into a message object.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
```diff
@@ -0,0 +1,189 @@
...
+import java.io.InputStream
+import java.io.OutputStream
+import java.nio.ByteBuffer
+import java.util.concurrent.ArrayBlockingQueue
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.Executors
+
+class StreamEnclaveHostInterface(
+        private val outputStream: OutputStream,     // Messages going to the host
+        private val inputStream: InputStream,       // Messages arriving from the host
```
shamsasari, line 15: What's the reasoning for abstracting away the socket/port and working at the stream level? It seems whever we will use these classes we will need to write the same networking code. It's also not clear which side is the server and which side is the client.

I suggest simplifying and just deal directly with ports. Have both this class and its counterpart take in a port, and have the host side start up the socket server and the enclave connect to it (all on localhost). There's no need to have any retries at this stage. We can assume a stable localhost connection for now. (I think going forward for prod use we will use shared memory for the IPC).

Related to this, I would rename this class to `EnclaveToHostSocketInterface` or just `HostSocketInterface`. And in that vein `EnclaveHostInterface` should be renamed likewise.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/StreamHostEnclaveInterface.kt
```diff
@@ -0,0 +1,218 @@
...
+ * This class is a streaming implementation of the [HostEnclaveInterface].
+ * It has three jobs:
+ *  - Serve as the endpoint for calls to make to the enclave, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Route calls from the enclave to the appropriate host side call handler, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Handle the low-level details of the messaging protocol (streamed ecalls and ocalls).
+ */
+class StreamHostEnclaveInterface(
+        private val outputStream: OutputStream,     // Messages going to the enclave
+        private val inputStream: InputStream,       // Messages arriving from the enclave
+) : HostEnclaveInterface() {
```
shamsasari, line 23: `HostToEnclaveInterface` or just `EnclaveInterface`

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /** Start the message receive loop thread. */
+    private val receiveLoopThread = Thread(receiveLoop, "Enclave message receive loop").apply { start() }
+
+    /**
+     * The shutdown process starts on the host.
+     * This function just blocks until the host sends a stop message and the message receive loop terminates.
+     */
+    fun awaitStop() {
```
shamsasari, line 64: Both classes should instead be implementing `Closeable`

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
```diff
@@ -0,0 +1,189 @@
...
+        outputStream.writeInt(maximumConcurrentCalls)
+    }
+
+    private val receiveLoop = object : Runnable {
+        private var done = false
+
+        /** Receive messages in a loop and send them to the appropriate call context. */
+        override fun run() {
+            while (!done) {
+                when (StreamCallInterfaceSignal.fromByte(inputStream.read().toByte())) {
```
shamsasari, line 31: `read` will return -1 on end of stream

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
```diff
@@ -0,0 +1,189 @@
...
+     */
+    fun awaitStop() {
+        receiveLoopThread.join()
+    }
+
+    /** Send a message to the receiving thread in the enclave-host interface. */
+    private fun sendMessageToHost(message: StreamCallInterfaceMessage) {
+        synchronized(outputStream) {
+            outputStream.write(StreamCallInterfaceSignal.MESSAGE.toByte().toInt())
+            message.writeToStream(outputStream)
```
shamsasari, line 72: Add a call to `flush`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
```diff
@@ -0,0 +1,189 @@
...
+                callExecutor.execute(newCallContext)
+                newCallContext
+            }
+            callContext.enqueMessage(message)
+        }
+
+        /** The host has told us there will be no more messages, shut everything down. */
+        private fun handleStopCommand() {
+            callExecutor.shutdown()     // Finish processing any outstanding calls
+            sendStopSignalToHost()     // Let the host know there will be no more messages
```
shamsasari, line 52: Is this an ack back to the host?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
shamsasari, line 15: rename these `toHost` and `fromHost` respectively. Likewise for the other class.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
```diff
@@ -0,0 +1,189 @@
...
+            done = true                 // Terminate the loop
+        }
+    }
+
+    /** Start the message receive loop thread. */
+    private val receiveLoopThread = Thread(receiveLoop, "Enclave message receive loop").apply { start() }
```
shamsasari, line 58: It's not good practice to start inside the c'tor. Have a separate start method

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/StreamHostEnclaveInterface.kt
```diff
@@ -0,0 +1,218 @@
...
+    private var isRunning = true
+
+    /** Receive the number of concurrent calls from the enclave. */
+    private val maxConcurrentCalls = inputStream.readInt()
+    private val callGuardSemaphore = Semaphore(maxConcurrentCalls)
+
+    private val receiveLoop = object : Runnable {
```
shamsasari, line 30: Why do we need to receive messages from the enclave on a background thread? Since the enclave never calls the host outside the context of an host->enclave call, we can simply block on the input stream waiting for a response from the enclave after the host has sent its message. If I understand the code correctly, that on its own would reduce a lot of the complexity in this class.

shamsasari (changes requested):

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
ac101m: Nope, it's the command that stops the receive loop on the host side. It might seem odd that it works this way, but I've learned from past mistakes that this kind of thing works best when the stop signal comes down the same channel that whatever async process you are trying to stop blocks on (rather than using `interrupt()` or any similar mechanism).

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/StreamCallInterfaceMessage.kt
ac101m: I didn't put much thought into the serialisation of the message data object! Will remove this in favour of `Data(Input|Output)Stream`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
ac101m: The idea was to keep the implementation agnostic of the stream implementation, so that we can use other stream implementations if we like (named pipes etc). The intention would be to add a `Socket(HostEnclave|EnclaveHost)Interface` class that is either a subclass of this one, or has an instance of this one as a member.

I also strongly disagree with the `EnclaveToHost` or `HostToEnclave` naming idea (marco suggested something similar) as these classes are bidirectional and handle incoming calls as well as outgoing ones. For reference `HostEnclaveInterface` is supposed to read as "the hosts interface to the enclave". If you can think of a name that better expresses this, I'm all for it!

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/StreamHostEnclaveInterface.kt
ac101m: Multiple threads need to be able to call the enclave simultaneously!

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/StreamCallInterfaceMessage.kt
ac101m: I tried this and it's actually substantially slower! (at least in the hammer tests that I've written)

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
ac101m: I've removed the StreamCallInterfaceSignal class and just put it all in the message class.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
ac101m: Because the host receive thread is blocking on the input stream from the enclave, so it makes sense that the sop command come from there.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/StreamHostEnclaveInterface.kt
shamsasari: But you can't send messages down the socket stream simultaneously, as evidenced by your locking of the streams. So we might as well simplify things and block on the receiving stream, waiting for the enclave's response. This is another reason why the socket implementation is probably temporary and we will need to look into shared memory for example at some point.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
shamsasari: Drop "Enclave" from the name then. It makes the name more unwieldy and context should make it clear where the source is (since it's in the enclave mode). Then the KDoc can make explicity the source and target.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/StreamHostEnclaveInterface.kt
ac101m: As per our conversation, I've added a comment to the send methods.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
ac101m: Fixed

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/StreamEnclaveHostInterface.kt
ac101m: Will leave as is for now due to large numbers of code changes required.

shamsasari:
@bonifazi-r3 please also take a look.

shamsasari:
Added Marco as a reviewer to get more exposure on this code.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    /** Shut down both the host and the enclave call interface. */
+    override fun close() {
+        /**
+         * Prevent new calls from entering the interface. This has to be synchronized to avoid race condition between
+         * the semaphore acquisition and the lockout check in [executeOutgoingCall].
+         */
+        synchronized(state) {
+            if (state == State.STOPPED) return
+            check(state == State.RUNNING) { "Call interface is not running." }
```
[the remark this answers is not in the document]
ac101m: Done. Resolve this conversation if you're happy with it.

ac101m:
> Added Marco as a reviewer to get more exposure on this code.

@bonifazi-r3 Just a heads up, I wouldn't spend _too_ much time looking at this. The performance is good enough for a PoC but it's unlikely to be sufficient for the final thing, we're probably going to need to replace it later. I'd advise you just familiarise yourself with the constructors and the start/stop methods.

bonifazi-r3:
> > Added Marco as a reviewer to get more exposure on this code.
> 
> @bonifazi-r3 Just a heads up, I wouldn't spend _too_ much time looking at this. The performance is good enough for a PoC but it's unlikely to be sufficient for the final thing, we're probably going to need to replace it later. I'd advise you just familiarise yourself with the constructors and the start/stop methods.

Having a look now and we can have a quick chat tomorrow morning if I have questions, deal?

shamsasari:
Yeah, don't spend too much time on this. More so that you get familiar with the code and it makes sense.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
```diff
@@ -0,0 +1,239 @@
...
+
+        fun sendMessage(messageType: SocketCallInterfaceMessageType, callTypeID: Byte, payload: ByteBuffer?) {
+            val outgoingMessage = SocketCallInterfaceMessage(
+                    hostThreadID, messageType, callTypeID, payload?.getAllBytes(avoidCopying = true))
+
+            sendMessageToHost(outgoingMessage)
+        }
+
+        fun sendCallMessage(callType: HostCallType, parameterBuffer: ByteBuffer?) {
+            requireNotNull(parameterBuffer)
```
shamsasari, line 148: The parameter buffer doesn't need to be null it seems

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
```diff
@@ -0,0 +1,239 @@
...
+            sendMessage(SocketCallInterfaceMessageType.CALL, callType.toByte(), parameterBuffer)
+        }
+
+        fun sendReturnMessage(callType: EnclaveCallType, returnBytes: ByteBuffer?) {
+            sendMessage(SocketCallInterfaceMessageType.RETURN, callType.toByte(), returnBytes)
+        }
+
+        fun sendExceptionMessage(callType: EnclaveCallType, exceptionBuffer: ByteBuffer?) {
+            requireNotNull(exceptionBuffer)
```
shamsasari, line 157: likewise

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
```diff
@@ -0,0 +1,237 @@
...
+            if (replyMessageType == SocketCallInterfaceMessageType.EXCEPTION) {
+                checkNotNull(replyPayload) { "Received exception message without parameter bytes." }
+                throw ThrowableSerialisation.deserialise(replyPayload)
+            }
+
+            return replyPayload?.let { ByteBuffer.wrap(replyPayload) }
+        }
+
+        /** Handle the initial call */
+        override fun run() {
```
shamsasari, line 204: If this is only run once, then it's better to not have the `EnclaveCallContext` implement `Runnable`. Otherwise it gives the impression this run method is run periodically or is the main focus of the class. I would suggest rename this to `initialRun` or something and just invoke in on the executor like `newCallContext::initialRun` or similar.

PR #60, opened 2022-10-13
CON-1160: Replace Gramine env variable with Gradle config flag

bonifazi-r3:
This is a first step in the direction of integrating Gramine into Conclave. Here we add the related flag in the `build.gradle` files of the enclaves and we allow the `EnclaveHost` to start Gramine enclaves, albeit with some assumptions/dummy behaviour.
The changes should have no impacts on the non-Gramine code.

In order to test a Gramine based enclave (only with a bash script now), I have run a "Hello World" example and checked that a process was created on my Linux machine (in addition to the temp directories).


plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
```diff
@@ -18,7 +18,11 @@ abstract class ConclaveTask : DefaultTask() {
     // workingDir: Set the working directory for the command. The defaults to the project directory
     // ignoreExitValue: If set to false, exceptions are raised if the command returns a value different from zero. Otherwise, please check the result.exitValue to check if the command was successful
     // errorOutputStream: Set the error output stream if you are interested in capturing the error messages generated by the command. If set to null, the error messages are written to standard error
-    data class CommandLineConfig(val workingDir: String? = null, val ignoreExitValue: Boolean = false, val errorOutputStream: ByteArrayOutputStream? = null)
+    data class CommandLineConfig(
```
bonifazi-r3, line 21: This class had few indentation issues.
There are no functional changes.

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -1041,6 +1056,8 @@ class EnclaveHost private constructor(
             val results = ArrayList<ScanResult>()
             EnclaveMode.values().mapNotNullTo(results) { findNativeEnclave(enclaveClassName, it) }
             findMockEnclave(enclaveClassName)?.let { results += it }
+            EnclaveMode.values()
```
shamsasari, line 1059: Have a single loop which does both graalvm above and gramine here.

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -1050,6 +1067,19 @@ class EnclaveHost private constructor(
...
+            //  As an example, if enclaveClassName is "com.r3.MyEnclave" and enclaveMode is "simulation",
+            //    we expect to find the manifest and the jar file under the directory "/com/r3/MyEnclave-simulation"
+            val filesLocation = "/${enclaveClassName.replace('.', '/')}-${enclaveMode.name.lowercase()}"
+            val manifestUrl = EnclaveHost::class.java.getResource("$filesLocation/${GRAMINE_ENCLAVE_MANIFEST}")
+            val jarUrl = EnclaveHost::class.java.getResource("$filesLocation/${GRAMINE_ENCLAVE_JAR_NAME}")
+
+            return if (manifestUrl == null || jarUrl == null) {
+                null
+            } else
+                ScanResult.Gramine(enclaveClassName, enclaveMode, manifestUrl, jarUrl)
```
shamsasari, line 1080: braces for this as well

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -1091,6 +1121,32 @@ class EnclaveHost private constructor(
...
+         * Performs a search for Gramine enclave files using ClassGraph. The search is performed by looking for 
+         * the manifest file and assuming that the other required files are also present there.
+         */
+        private fun findGramineEnclaves(classGraph: ClassGraph, results: MutableList<ScanResult>) {
+            //  As an example, if we find the manifest file is in the directory "com/r3/MyEnclave-simulation"
+            //    the resulting enclaveClassName will be "MyEnclave", enclaveMode will be "simulation" and
+
+            classGraph.scan().use {
+                for (resource in it.allResources) {
+                    val manifestSearchPattern = Pattern.compile("""^(.+)-(simulation|debug|release)/(bash\.manifest)$""")
```
shamsasari, line 1134: Move this out of the loop

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -1122,7 +1178,19 @@ class EnclaveHost private constructor(
                 init {
                     require(enclaveMode != EnclaveMode.MOCK)
                 }
-                override fun toString(): String = "${enclaveMode.name.lowercase()} $enclaveClassName"
+                override fun toString(): String = "graal ${enclaveMode.name.lowercase()} $enclaveClassName"
```
shamsasari, line 1181: graalvm, and rename this class as well

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
```diff
@@ -0,0 +1,111 @@
...
+            enclaveDirectory.toFile().deleteRecursively()
+        } catch (e: IOException) {
+            logger.debug("Unable to delete temp directory $enclaveDirectory", e)
+        }
+    }
+
+    private fun copyGramineFilesToWorkingDirectory() {
+        //  Here we copy files from inside the jar into a temporary folder
+        val manifestName = File(manifestUrl.file).name
+        val jarName = File(jarUrl.file).name
```
shamsasari, line 97: Do`Paths.get(url.toURI())` instead

shamsasari: However, aren't the manifest and jar names already known, i.e. `GRAMINE_ENCLAVE_JAR_NAME` and  `GRAMINE_ENCLAVE_MANIFEST`?

shamsasari (changes requested):
I had a feeling the plugin changes would be too complex to review. Can you extract that cleanup into a seperate PR first. Thanks.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineHostEnclaveInterface.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * This class is the implementation of the [HostEnclaveInterface] for Gramine enclaves.
+ * It is currently in an experimental state. Do not use it.
+ * //  TODO: Refactor this class to move it to a non experimental state
+ *  - Serve as the endpoint for calls to make to the enclave, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Route calls from the enclave to the appropriate host side call handler, see [com.r3.conclave.common.internal.CallInterface]
+ *  - Handle the low-level details of the messaging protocol (ECalls and OCalls).
+ */
+class GramineHostEnclaveInterface(private val connector: MockCallInterfaceConnector) : HostEnclaveInterface() {
```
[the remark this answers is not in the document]
bonifazi-r3: This class is used by `GramineEnclaveHandle`, just to provide a default/dummy interface. This will be refactored by @ac101m  I guess. I needed it so that the Gramine enclave does  not crash when starting and we avoid a lot of conditional code

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
bonifazi-r3: Done

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/RuntimeType.kt
```diff
@@ -0,0 +1,6 @@
+package com.r3.conclave.plugin.enclave.gradle
+
+enum class RuntimeType {
+    Gramine,
+    Graal
```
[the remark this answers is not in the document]
bonifazi-r3: The previous flag was `graalvm-native-image` I think.
Going to use `graalvm` (GraalVM in the enum here).

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -1018,6 +1032,7 @@ class EnclaveHost private constructor(
             val classGraph = ClassGraph()
             val results = ArrayList<ScanResult>()
             findNativeEnclaves(classGraph, results)
```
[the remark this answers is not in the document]
bonifazi-r3: DOne

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
bonifazi-r3: Done

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
bonifazi-r3: Great, thanks

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
bonifazi-r3: It should be done

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -229,6 +231,20 @@ class EnclaveHost private constructor(
...
+        // The internal modifier prevents this from appearing in the API docs, however because we shade Kotlin it will
+        // still be available to Java users. We solve that by making it synthetic which hides it from the Java compiler.
+        @JvmSynthetic
+        @JvmStatic
+        internal fun internalCreateGramine(
+            enclaveMode: EnclaveMode,
+            enclaveClassName: String,
+            manifest: URL,
+            jar: URL
+            ): EnclaveHost {
```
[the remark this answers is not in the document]
bonifazi-r3: Sorry, done

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -123,6 +126,60 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+
+    inner class IntoGramineTask(
+        private val enclaveClassNameTask: EnclaveClassName,
+        private val typeLowerCase: String
+    ) : Action<Task> {
+        override fun execute(task: Task) {
+            //  Note that in Gramine we use the class name as a folder,
+            //     not as a part of a file, as it is done with Native tasks
+            val location = enclaveClassNameTask.outputEnclaveClassName.get().replace('.', '/') + "-${typeLowerCase}"
+            println("location $location")
```
shamsasari, line 177: delete

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -427,7 +496,19 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+            target.afterEvaluate {
+                val runtimeType = conclaveExtension.runtime.get()
+
+                when (runtimeType) {
+                    RuntimeType.Gramine -> {
+                        target.artifacts.add(typeLowerCase, buildGramineEnclaveJarTask.get().archiveFile)
+                    }
+
+                    RuntimeType.GraalVM -> {
+                        target.artifacts.add(typeLowerCase, signedEnclaveJarTask.get().archiveFile)
```
shamsasari, line 508: rename `signedEnclaveJarTask` to incorporate graalvm

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/BuildUnsignedGramineEnclave.kt
```diff
@@ -0,0 +1,65 @@
...
+            logger.info("Copying $MANIFEST_TEMPLATE to build directory")
+            Files.copy(it, outputManifestTemplateFile, StandardCopyOption.REPLACE_EXISTING)
+        }
+        return outputManifestTemplateFile
+    }
+
+    private fun generateGramineDirectManifest(templateManifest: String) {
+        val command = listOf(
+            GRAMINE_MANIFEST_EXECUTABLE,
+            "-Dlog_level=${logType.get().type}",
```
shamsasari, line 57: Rather than creating a seperate log level for gramine, instead link it to gradle's logging level. So if we specify info or debug level in gradle then gramine will also do the same. This we can get configure gramine's logging level with no effort, but rather here the logging level has been hardcoded in the plugin class.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
shamsasari: Actually no. Can you instead update `signedEnclaveJarTask` so that it does either the graalvm path or the gramine path. Because this task is what links the enclave module to the host one, i'd rather not have two different types and have a single gradle task for it.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/BuildUnsignedGramineEnclave.kt
bonifazi-r3: I can link Gramine log level to Gradle one with a mapping, but my idea is that this needs to be hardcoded in the template manifest and set only to error.
I don't think we expect the manifest generation to fail often, i.e. we should make it so generic that it does never fail. I see the Gramine log something that we do not want to give the user, i.e. I guess we will need to capture and rethrow the exceptions with a nicer message.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
bonifazi-r3: Mmm, should we use "signed" for Gramine as well? I.e. `gramine-direct` does not require signing as far as I understand.
Maybe we need an extra "final" task?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/BuildUnsignedGramineEnclave.kt
shamsasari: In that case I don't see any benefit to the Gramine Log Level class, and should be deleted.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
shamsasari: For now use this "signed" task. I think we need to rename it anyway at some point, but for now it's fine.

bonifazi-r3: Is there any reason for using one task only at the end actually? I.e. why not be explicit and clear, i.e. separating the graphs for Gramine and Graal? I am trying to implement it and I notice that the code would become a bit conditional and ugly (changing the task after evaluation), and the user would actually risk more confusion by seeing the same task both for Gramine or Graal cases.

shamsasari: Because the output of the enclave build process be encapsulated by a single task regardless of what the enclave config is. It makes it simpler to reason about, for us as the Conclave platform devs. This task isn't really exposed to the end user so they won't be confused, but I also don't think they would be confused anyway. In fact, I think the opposite of having separate tasks would be more confusing.

bonifazi-r3: Ok. I am trying to do that, but it looks to me that we are by purpose forcing some Gradle mechanism. How can we make `dependsOn` conditional and also make some of its dependencies conditional at runtime nicely?
My understanding of the Gradle mechanism is that we have that configuration and then execution phase. It really seems to me as if we actually need to create other 3 configurations and create the graph for them at compile/configuration time, and then execute the configuration according to the runtime selection.
I.e. I see runtimeType very similar to enclaveMode in the context of running these tasks.

Instead we seem to want to create our own graph at runtime. Is this really how Gradle is meant to behave? Is there a convention around that?

It seems to me that if we remove the "constraint" of using the same name of the last task, but instead we add the runtime type to all the tasks (as we do for the buildType) , our life would be easier and the logic would be straightforward.

bonifazi-r3: I see that this would require a change to the build directory, etc. And it's kind of messy to apply this now, as I expect a lot of code to have issues with it. Not sure what to do, I am tempted just to rename the final Gramine task with a "signed" naming in it, to make it similar to the previous one (and maybe rename the Graal one with "GraalVM" in it.
Any suggestion? Or do you know how to make dependencies conditional?

shamsasari: OK no worries, let's leave this as is for now. We can readdress this later.

PR #61, opened 2022-10-14
CON-851: Reduce configurations in reflection files after Graal updates in version 21.3

bonifazi-r3:
Graal 21.3 improvements allow to have smaller configuration files used for reflection, in this patch we make use of it.


shamsasari (approved):
Thanks for this.

PR #63, opened 2022-10-18
CON-1126 Version bump on dependencies excluding Kotlin and Gradle

filipesoliveira:
**Note:**
`locationtech_jts_version` and `h2database_h2_version` have not been updated to the next version because they will cause the integration test to fail.


shamsasari (approved):

filipesoliveira:
locationtech_jts_version and h2database_h2_version were already in the most recent version which does not break the integration tests.

PR #66, opened 2022-10-18
[CON-1172] Address un-addressed review comments

ac101m:
# Changes
- Use state manager class for host enclave interface rather than plain enum.

### DIFF
--- a/conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
+++ b/conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
@@ -73,25 +73,28 @@ class SocketHostEnclaveInterface(private val port: Int) : HostEnclaveInterface()
     fun start() {
         synchronized(stateManager) {
             if (stateManager.state == State.Running) return
-            stateManager.checkStateIs<State.Ready> {
+            stateManager.transitionStateFrom<State.Ready>(to = State.Running) {
                 "Interface may not be started multiple times."
             }
 
-            /** Start the socket and instantiate the data input/output streams. */
-            ServerSocket(port).use {
-                socket = it.accept()
-                toEnclave = DataOutputStream(socket.getOutputStream())
-                fromEnclave = DataInputStream(socket.getInputStream())
+            try {
+                /** Start the socket and instantiate the data input/output streams. */
+                ServerSocket(port).use {
+                    socket = it.accept()
+                    toEnclave = DataOutputStream(socket.getOutputStream())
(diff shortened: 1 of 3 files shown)


ac101m:
Some changes which I failed to commit for the original PR.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -88,7 +90,8 @@ class SocketHostEnclaveInterface(private val port: Int) : HostEnclaveInterface()
             /** Start the message receive thread and allow calls to enter. */
             receiveLoopThread.start()
             callGuardSemaphore.release(maxConcurrentCalls)
-            state = State.RUNNING
+
+            stateManager.transitionStateFrom<State.Ready>(to = State.Running)
```
shamsasari, line 94: You can move this up and replace it with the `checkStateIs`.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -109,7 +114,7 @@ class SocketHostEnclaveInterface(private val port: Int) : HostEnclaveInterface()
             sendMessageToEnclave(SocketCallInterfaceMessage.STOP_MESSAGE)
             receiveLoopThread.join()
 
-            state = State.STOPPED
+            stateManager.transitionStateFrom<State.Running>(to = State.Stopped)
```
shamsasari, line 117: ditto

shamsasari (commented):
I guess there's also the "Interface" to "Link" rename?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
ac101m: Is this wise? The operations in between can fail. Let me know if I'm over-thinking it.

ac101m:
> I guess there's also the "Interface" to "Link" rename?

You want to do that now? It will involve a large number of changes. Comment adjustments, etc.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
shamsasari: if `start()` fails then you don't want it to be possible to call it again.

shamsasari (approved):

PR #67, opened 2022-10-18
CON-1175: Start a basic java Gramine non-SGX process from EnclaveHost.start

anien:
Create GramineEntryPoint in conclave-enclave. Specify this entrypoint when running java gramine-direct process.
Rename bash manifest to java manifest and change manifest entrypoint to java.


shamsasari (commented):
I'll leave it to Marco to review, but looks good to me. Thanks.

PR #71, opened 2022-10-19
[CON-1177] Wire up new host-enclave socket connection to existing handlers

ac101m:
# Changes
- Bind port at host interface creation rather than when `start()` is called.
- Improve performance of socket call interface .
- Implement `GrameneEnclaveEnvironment` with mock attestation.


shamsasari (changes requested):
We need to discuss integration tests. One way could be to parameterise the `runtime` flag for all the enclave projects in `integration-tests` based on a Gradle property. That way we would easily parallelise the run test on TC with separate Gramine and GraalVM jobs. Obviously we would uncover a lot of failing tests, but it might be something to make optional and we would have an idea of what's missing and what's still left to do.

Or we could just have some bespoke tests for now to test the basic integration. But they would be temporary.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -1,12 +1,72 @@
...
+
+        /** The host passes the port on the command line. */
+        private fun getPortFromArgs(args: Array<String>): Int {
+            if (args.isEmpty()) {
+                System.err.println(USAGE_STRING)
+                exitProcess(EXIT_ERR)
+            }
+
+            return try {
+                args[0].toUInt().toInt()
```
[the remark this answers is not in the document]
ac101m: toUint() will presumably throw if the input is negative? That's what I was going for. Should probably use toUShort() actually now that I think of it...

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
```diff
@@ -82,10 +88,12 @@ class SocketEnclaveHostInterface(
             }
 
             try {
-                /** Connect the socket and instantiate data input/output streams. */
-                socket = Socket(host, port)
-                toHost = DataOutputStream(socket.getOutputStream())
-                fromHost = DataInputStream(socket.getInputStream())
+                /** Connect sockets and instantiate data input/output streams. */
+                fromHostSocket = Socket(host, port).apply { tcpNoDelay = true }
+                toHostSocket = Socket(host, port).apply { tcpNoDelay = true }
```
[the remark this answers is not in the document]
ac101m: Considering that I got a considerable performance boost when I did this, I'd say not!

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -1138,7 +1133,7 @@ class EnclaveHost private constructor(
                     val pathMatcher = manifestSearchPattern.matcher(resource.url.path)
 
                     if (pathMatcher.matches()) {
-                        val enclaveClassName = pathMatcher.group(1).replace('/', '.')
+                        val enclaveClassName = pathMatcher.group(1).replace('/', '.').substringAfter("!.")
```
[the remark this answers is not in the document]
ac101m: The string that comes out of classgraph is prefixed with the absolute path of the jar, but we just want the fqcn. This just removes the path.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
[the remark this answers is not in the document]
shamsasari: Can you add an explanatory comment.

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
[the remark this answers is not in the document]
shamsasari: Can you give an example. Try to understand what's different to cause this change in behaviour.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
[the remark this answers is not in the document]
ac101m: I've cleaned this up a bit. More verbose but less confusing now.

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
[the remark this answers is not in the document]
ac101m: So the result from classgraph looks something like this: `<absolute path to shadow jar>!.<enclave fqcn>`. I'm not sure why this differs from native mode to be honest.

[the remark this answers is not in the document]
ac101m: Figured this out, substring call is no longer needed.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -0,0 +1,118 @@
...
+        if (reportData != null) {
+            body[SgxReportBody.reportData] = reportData.buffer
+        }
+        body[SgxReportBody.cpuSvn] = ByteBuffer.wrap(currentCpuSvn)
+        body[SgxReportBody.mrenclave] = ByteBuffer.wrap(mrenclave)
+        body[SgxReportBody.mrsigner] = ByteBuffer.wrap(mrsigner)
+        body[SgxReportBody.isvProdId] = productID
+        // Revocation level in the report is 1 based. We subtract 1 from it when reading it back from the report.
+        body[SgxReportBody.isvSvn] = revocationLevel + 1
+        body[SgxReportBody.attributes][SgxAttributes.flags] = SgxEnclaveFlags.DEBUG
```
bonifazi-r3, line 61: Do you know why is this "DEBUG"?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -21,10 +21,15 @@ import java.util.concurrent.Semaphore
  *  - Route calls from the enclave to the appropriate host side call handler, see [com.r3.conclave.common.internal.CallInterface]
  *  - Handle the low-level details of the messaging protocol (socket with streamed ECalls and OCalls).
  */
-class SocketHostEnclaveInterface(private val port: Int) : HostEnclaveInterface(), Closeable {
-    private lateinit var socket: Socket
+class SocketHostEnclaveInterface(port: Int = 0, private val maxConcurrentCalls: Int) : HostEnclaveInterface(), Closeable {
```
bonifazi-r3, line 24: Is actually defaulting to port 0 nice?

bonifazi-r3 (changes requested):

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
shamsasari: No particular reason. Just seems to make more sense to have it than not.

shamsasari:
What's the story on integration tests?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
```diff
@@ -1,89 +1,91 @@
...
+        /** Start the local call interface. */
+        val interfaceStartTask = FutureTask { enclaveInterface.start() }
+        val interfaceStartThread = Thread(interfaceStartTask).apply { start() }
+
+        try {
+            /** Start the enclave process, passing the port that the call interface is listening on. */
+            processGramineDirect = ProcessBuilder()
+                .inheritIO()
+                .directory(enclaveDirectory.toFile())
+                .command("gramine-direct", "java", "-cp", GRAMINE_ENCLAVE_JAR_NAME, "com.r3.conclave.enclave.internal.GramineEntryPoint", enclaveInterface.port.toString())
```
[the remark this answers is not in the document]
ac101m: I don't think having that line in the manifest is necessarily bad, provided the code that handles the command line parameters (`GrameneEntryPoint`) is safe. Also, any form of inter-process communication is going to require that the enclave know how to address the host or vice versa anyway, so I'm not sure there is a way to avoid having to pass _something_ in an un-trusted manner be it by file or by using command line parameters.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
ac101m: Port zero means "system, please allocate me a free port"

ac101m:
> What's the story on integration tests?

Haven't touched them. I can add a flag, but I'd wager the vast majority will fail at the moment.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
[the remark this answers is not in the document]
bonifazi-r3: My point is "do we want to make the enclave-side port part of the enclave trusted codebase"? Not sure, I will think about this. I feel that there might be some risk if we allow the same enclave to run the same secure enclave on different ports for example, but can't think of a specific case right now.
Also, as things stand, only the `java` command is actually signed to trust under Gramine, i.e. attackers can really run whatever the `java` command allows them to do under Gramine, for whatever reason.
It is safer if even the `-cp JARNAME ...` part is covered by the signing step and this is possible by using the suggested mechanisms in the docs
```
loader.argv = ["arg0", "arg1", "arg2", ...]

or

loader.argv_src_file = "file:file_with_serialized_argv"
.
```

Finally, I would not really leave a line of the manifest where it says `__insecure`. We will need to explain and discuss and I expect this to be annoying. Of course, if this is our only possibility we will need to provide plenty of explanations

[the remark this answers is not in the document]
ac101m: I've added a TODO regarding sensitisation of the command line parameters.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    override fun sealData(toBeSealed: PlaintextAndEnvelope): ByteArray {
+        return EnclaveUtils.sealData(aesSealingKey, toBeSealed)
+    }
+
+    override fun unsealData(sealedBlob: ByteBuffer): PlaintextAndEnvelope {
+        return EnclaveUtils.unsealData(aesSealingKey, sealedBlob)
+    }
+
+    // TODO: (CON-1178) Implement proper attestation
```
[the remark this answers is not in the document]
ac101m: I've cleaned up the TODOs in here so they point at the new ticket from this morning.

integration-tests/build.gradle
```diff
@@ -5,6 +5,7 @@ subprojects {
     apply plugin: 'java'
 
     ext.enclaveMode = findProperty("enclaveMode")?.toString()?.toLowerCase() ?: "simulation"
+    ext.runtimeType = findProperty("runtimeType")?.toString() ?: "GraalVM"
```
shamsasari, line 8: Is this particular variable used elsewhere?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
```diff
@@ -25,9 +26,7 @@ open class ConclaveExtension @Inject constructor(objects: ObjectFactory) {
     val reflectionConfigurationFiles: ConfigurableFileCollection = objects.fileCollection()
     val serializationConfigurationFiles: ConfigurableFileCollection = objects.fileCollection()
 
-    val runtime: Property<RuntimeType> = objects.property(RuntimeType::class.java).convention(RuntimeType.GraalVM)
-    val graalvm = RuntimeType.GraalVM
-    val gramine = RuntimeType.Gramine
+    val runtime: Property<String> = objects.property(String::class.java).convention(RuntimeType.GraalVM.name)
```
shamsasari, line 29: Why change this to a String? Doesn't Gradle handle the parsing for you?

shamsasari:
You'll need to update ci_integration_tests.sh as well to be able to create a separate job on TC.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
ac101m: Not that I could tell. The two lines above here (`val graalvm = RuntimeType.GraalVM` and `val gramine = RuntimeType.Gramine`) are what was being referenced by gradle when we specified `runtime = <whatever>`. So no parsing at all so far as I can tell. If you tried to specify something other than those, you got the following error:

```
A problem occurred evaluating project ':enclave'.
> Could not get unknown property 'gramie' for extension 'conclave' of type com.r3.conclave.plugin.enclave.gradle.ConclaveExtension.
```

integration-tests/build.gradle
ac101m: > You'll need to update ci_integration_tests.sh as well to be able to create a separate job on TC.

Script & TC updated.

ac101m:
> You'll need to update ci_integration_tests.sh as well to be able to create a separate job on TC.

Script & TC updated.

integration-tests/build.gradle
ac101m: I just copied the line above without thinking too much. Does this need changing?

shamsasari: I believe you can change all the `findProperty("runtimeType")` to something like `ext.runtimeType`

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
shamsasari: Perhaps add a comment so that someone doesn't revert this back to an enum

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -493,16 +502,17 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+                    when (RuntimeType.valueOf(conclaveExtension.runtime.get())) {
+                        RuntimeType.Gramine -> {
+                            target.artifacts.add(typeLowerCase, signedEnclaveGramineJarTask.get().archiveFile)
+                        }
+                        RuntimeType.GraalVM -> {
+                            target.artifacts.add(typeLowerCase, signedEnclaveJarTask.get().archiveFile)
+                        }
                     }
+                } catch (e: Exception) {
+                    /** Do nothing. Error condition is handled in [apply]. */
```
shamsasari, line 515: What happens if you don't have this catch?

shamsasari:
Duplicate the Simulation Integration Test job and modify it for Gramine. But before you do that, you'll need to update the test block in integration-tests/build.gradle to not fail fast if in gramine mode. The default to failfast is set in the common.gradle and so we'll need to override that.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
ac101m: The exception is printed an additional three times, once for each simulation, debug and release.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
ac101m: Done

integration-tests/build.gradle
ac101m: Yeah, not sure why I did this. Fixed.

shamsasari (approved):

shamsasari:
@bonifazi-r3 please take a look .

integration-tests/build.gradle
bonifazi-r3, line 8: Can we guarantee that `graalvm` is the same as `GraalVm` or `GraalVM`, etc?

bonifazi-r3 (changes requested):

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -75,6 +76,14 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
                         "Please set the 'revocationLevel' property in the build configuration for your enclave.\n" +
                         "If you're unsure what this error message means, please consult the conclave documentation.")
             }
+            // Check the passed runtime type is valid.
+            try {
+                RuntimeType.valueOf(conclaveExtension.runtime.get())
```
shamsasari, line 81: convert the string to uppercase to fix the issue highlighted by @bonifazi-r3 .

integration-tests/build.gradle
ac101m: At the moment it's strict, you have to use the exact capitalisation, but the error message contains the correct capitalisation if you get it wrong:

```
> 'gramine' is not a valid enclave runtime type.
  Valid runtime types are: [Gramine, GraalVM].
```

Can uppercase/lowercase this if you feel it's necessary.

bonifazi-r3: I'd personally keep consistent with the `enclaveMode` now and make it lowercase. If not, I expect time wasted in remembering if it was `GraalVM` or `GraalVm`

containers/integration-tests-build/Dockerfile
```diff
@@ -17,6 +17,12 @@ RUN wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.k
 RUN wget -qO - https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
 RUN add-apt-repository "deb [arch=amd64] https://packages.microsoft.com/ubuntu/20.04/prod focal main"
 
+# Install gramine for the in-kernel SGX driver
+RUN curl -fsSLo /usr/share/keyrings/gramine-keyring.gpg https://packages.gramineproject.io/gramine-keyring.gpg
+RUN echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/gramine-keyring.gpg] https://packages.gramineproject.io/ focal main' | tee /etc/apt/sources.list.d/gramine.list
+RUN apt update
```
shamsasari, line 24: Aren't we already doing an apt-get update above?

containers/integration-tests-build/Dockerfile
```diff
@@ -17,6 +17,12 @@ RUN wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.k
...
+RUN apt install gramine
```
shamsasari, line 25: apt vs apt-get?

shamsasari: Can you merge this into the install command below.

integration-tests/build.gradle
shamsasari: Let's make it case insensitive. It's easier for the user.

containers/integration-tests-build/Dockerfile
ac101m: Yup, we need to do it again after adding the gramine gpg key.

integration-tests/build.gradle
ac101m: Gotcha, will make it case insensitive

bonifazi-r3 (approved):

PR #72, opened 2022-10-20
CON-1095 - Simplify conclave-build container

filipesoliveira:


containers/conclave-build/Dockerfile
```diff
@@ -12,24 +12,11 @@ ENV JAVA_HOME /usr/lib/jvm/java-17-openjdk-amd64
```
shamsasari, line 12: Do you know what the `sed` command is doing above? It would be nice to get some comment on it.

containers/scripts/ci_build_publish_docker_images.sh
```diff
@@ -114,4 +121,7 @@ fi
 if [ -z "${DOCKER_IMAGE_CONCLAVE_BUILD:-}" ] || [ "${DOCKER_IMAGE_CONCLAVE_BUILD}" == "1" ]; then
   buildAndPublishContainerIfItDoesNotExist $container_image_conclave_build buildContainerConclaveBuild
 fi
+if [ -z "${DOCKER_IMAGE_CONCLAVE_BUILD_INTEGRATION_TESTS:-}" ] || [ "${DOCKER_IMAGE_CONCLAVE_BUILD_INTEGRATION_TESTS}" == "1" ]; then
```
shamsasari, line 124: These are env variables? Why is it useful to turn this off via this env variable?

shamsasari (changes requested):
Can you rename the container to just `integration-tests-build` or something. I think we can drop the "conclave-".

shamsasari:
I've swapped out Marco for Anna, as she has as MBP and I'd like her to test this out. Thanks.

containers/conclave-build/Dockerfile
filipesoliveira: It is setting the locale to US. I am guessing it was added to fix a strange issue when building Conclave.

Reference: 
1) locale - https://wiki.archlinux.org/title/locale

containers/scripts/ci_build_publish_docker_images.sh
filipesoliveira: To speed up testing. It is nice to be able to disable building the containers that are not needed without having to change the scripts.

anien:
I have built the Conclave and run the integration tests outside of the dev container successfully on this branch on my Mac machine. Waiting with approval if Filipe is going to change the name of the container or not.

filipesoliveira:
> Can you rename the c

Done.

shamsasari (approved):

PR #73, opened 2022-10-21
CON-1133, CON-1188: Make stubbed functions to return an error

bonifazi-r3:
Many Linux Posix functions were previously stubbed in such a way that the enclave would abort when these were called.
We now want to return an error instead, so that it is up to the JVM (whose bytecode gets compiled with Native Image) interacting with the Posix calls to handle errors.

This PR also adds tests for the symbolic and hard link (https://r3-cev.atlassian.net/browse/CON-861, https://r3-cev.atlassian.net/browse/CON-862)

### DIFF
--- a/cpp/jvm-enclave-common/stubs/dlfcn.cpp
+++ b/cpp/jvm-enclave-common/stubs/dlfcn.cpp
@@ -10,35 +10,35 @@
 
 using namespace std;
 
-//////////////////////////////////////////////////////////////////////////////
-// Stub functions to satisfy the linker
-STUB(dlclose);
-
 extern "C" {
 
-void *dlopen(const char *filename, int flags) {
-    enclave_trace("dlopen\n");
-    return nullptr;
-}
+    void *dlopen(const char *filename, int flags) {
+        enclave_trace("dlopen\n");
+        return nullptr;
+    }
 
-void *dlmopen(const char *filename, int flags) {
-    enclave_trace("dlmopen\n");
-    return nullptr;
-}
+    void *dlmopen(const char *filename, int flags) {
+        enclave_trace("dlmopen\n");
+        return nullptr;
+    }
 
-char *dlerror(){
-    enclave_trace("dlerror\n");
-    return nullptr;
-}
+    char *dlerror(){
+        enclave_trace("dlerror\n");
+        return nullptr;
+    }
 
-int dlinfo(void *handle, int request, void *info) {
-    enclave_trace("dlinfo\n");
-    return -1;
-}
+    int dlinfo(void *handle, int request, void *info) {
+        enclave_trace("dlinfo\n");
+        return -1;
+    }
 
-int dladdr(void* addr, Dl_info* info) {
-    enclave_trace("dladdr\n");
-    return 0;
-}
+    int dladdr(void* addr, Dl_info* info) {
+        enclave_trace("dladdr\n");
+        return 0;
+    }
 
+    int dlclose(void *handle) {
+        enclave_trace("dlclose\n");
+        return -1;
+    }
 }
--- a/cpp/jvm-enclave-common/stubs/ioctl.cpp
+++ b/cpp/jvm-enclave-common/stubs/ioctl.cpp
@@ -3,10 +3,11 @@
 //
 #include "vm_enclave_layer.h"
 
-//////////////////////////////////////////////////////////////////////////////
-// Stub functions to satisfy the linker
-STUB(ioctl);
-
 extern "C" {
 
+    int ioctl(int fd, unsigned long request, ...) {
+        enclave_trace("ioctl\n");
+        errno = EBADF;
+        return -1;
+    }
 }
--- a/cpp/jvm-enclave-common/stubs/ioctl.cpp
+++ b/cpp/jvm-enclave-common/stubs/ioctl.cpp
@@ -7,7 +7,7 @@ extern "C" {
 
     int ioctl(int fd, unsigned long request, ...) {
         enclave_trace("ioctl\n");
-        errno = EBADF;
+        errno = ENOSYS;
         return -1;
     }
 }
--- a/cpp/jvm-enclave-common/stubs/stdio.cpp
+++ b/cpp/jvm-enclave-common/stubs/stdio.cpp
@@ -153,6 +153,6 @@ extern "C" {
 
     char *fgets(char *s, int size, FILE *stream) {
         enclave_trace("fgets\n");
-        return NULL;
+        return nullptr;
     }
 }
--- a/cpp/jvm-enclave-common/stubs/stdio.cpp
+++ b/cpp/jvm-enclave-common/stubs/stdio.cpp
@@ -4,10 +4,6 @@
 #include "vm_enclave_layer.h"
 #include "file_manager.h"
 
-//////////////////////////////////////////////////////////////////////////////
-// Stub functions to satisfy the linker
-STUB(fgets);
-
 extern "C" {
 
     FILE *stdin = nullptr;
@@ -154,4 +150,9 @@ extern "C" {
         }
         return res;   
     }    
+
+    char *fgets(char *s, int size, FILE *stream) {
+        enclave_trace("fgets\n");
+        return NULL;
+    }
 }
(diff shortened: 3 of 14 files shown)


cpp/jvm-enclave-common/stubs/dlfcn.cpp
```diff
@@ -10,35 +10,35 @@
...
-    enclave_trace("dlinfo\n");
-    return -1;
-}
-
-int dladdr(void* addr, Dl_info* info) {
-    enclave_trace("dladdr\n");
-    return 0;
-}
-
+    void *dlopen(const char *filename, int flags) {
```
bonifazi-r3, line 15: Just indentation here

integration-tests/general/common/src/main/kotlin/com/r3/conclave/integrationtests/general/common/tasks/FileSystemActions.kt
```diff
@@ -65,6 +65,27 @@ class DeleteFile(private val path: String, private val nioApi: Boolean) : FileSy
     override fun resultSerializer(): KSerializer<Boolean> = Boolean.serializer()
 }
 
+@Serializable
+class CreateSymlink(private val filePath: String, private val symlinkPath: String) : FileSystemAction<Unit>() {
+    override fun run(context: EnclaveContext, isMail: Boolean) {
+        val file = File(filePath)
+        val symLink = File(symlinkPath)
```
shamsasari, line 72: Use `Paths.get`

integration-tests/general/tests/src/test/kotlin/com/r3/conclave/integrationtests/general/tests/EnclaveHostNativeTest.kt
```diff
@@ -83,4 +84,13 @@ class EnclaveHostNativeTest : AbstractEnclaveActionTest() {
             enclaveHost().mockEnclave
         }
     }
+
+    @Test
+    fun `create socket fails`() {
+        Assertions.assertThatThrownBy {
```
shamsasari, line 90: static import

integration-tests/general/tests/src/test/kotlin/com/r3/conclave/integrationtests/general/tests/EnclaveHostNativeTest.kt
```diff
@@ -83,4 +84,13 @@ class EnclaveHostNativeTest : AbstractEnclaveActionTest() {
...
+            callEnclave(CreateSocket())
+        }
+            .isInstanceOf(com.r3.conclave.common.EnclaveException::class.java)
+            .hasCauseExactlyInstanceOf(java.net.SocketException::class.java)
```
shamsasari, line 94: import

shamsasari:
A few comments, but @ac101m is better placed to review than me.

cpp/jvm-enclave-common/stubs/stdio.cpp
```diff
@@ -154,4 +150,9 @@ extern "C" {
         }
         return res;   
     }    
+
+    char *fgets(char *s, int size, FILE *stream) {
+        enclave_trace("fgets\n");
+        return NULL;
```
ac101m, line 156: nullptr

cpp/jvm-enclave-common/stubs/sys_poll.cpp
```diff
@@ -3,9 +3,12 @@
...
-//////////////////////////////////////////////////////////////////////////////
-// Stub functions to satisfy the linker
-STUB(poll);
-
 extern "C" {
+
+    typedef unsigned long int nfds_t;
+
+    int poll(struct pollfd *fds, nfds_t nfds, int timeout) {
+        errno = ENOMEM;
```
ac101m, line 11: Did some reading and it looks like we can just return ENOSYS from anything, even if it's not documented for that function.
These all defined so that they have different values, see here:
- https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/errno-base.h#L16
- https://elixir.bootlin.com/linux/latest/source/arch/alpha/include/uapi/asm/errno.h#L51

cpp/jvm-enclave-common/stubs/ioctl.cpp
```diff
@@ -3,10 +3,11 @@
...
 
-//////////////////////////////////////////////////////////////////////////////
-// Stub functions to satisfy the linker
-STUB(ioctl);
-
 extern "C" {
 
+    int ioctl(int fd, unsigned long request, ...) {
+        enclave_trace("ioctl\n");
+        errno = EBADF;
```
ac101m, line 10: ENOSYS

cpp/jvm-enclave-common/stubs/dlfcn.cpp
```diff
@@ -10,35 +10,35 @@
...
+
+    int dladdr(void* addr, Dl_info* info) {
+        enclave_trace("dladdr\n");
+        return 0;
+    }
+
+    int dlclose(void *handle) {
+        enclave_trace("dlclose\n");
+        return -1;
+    }
```
ac101m, line 43: No need to set errno on these?

cpp/jvm-enclave-common/stubs/sys_socket.cpp
```diff
@@ -3,26 +3,105 @@
...
 extern "C" {
 
     int socketpair(int domain, int type, int protocol, int sv[2]) {
-	return socketpair_impl(domain, type, protocol, sv);
+        return socketpair_impl(domain, type, protocol, sv);
+    }
+
+    int accept(int sockfd, struct sockaddr *addr, socklen_t *addrlen) {
+        enclave_trace("accept\n");
+        errno = EINVAL;
```
ac101m, line 14: ditto, and for all others

ac101m (commented):
Overall good, definitely the right direction. But I think we should probably use `ENOSYS` for most if not all of these! It seems as though the posix error codes are all defined such that their numeric values do not conflict.

cpp/jvm-enclave-common/stubs/stdio.cpp
bonifazi-r3: This is weird somehow, this is C code that gets compiled with C++ compiler. `man fgets` obviously does not mention `nullptr` but only `NULL`. For consistency with the remaining code we have, I would agree to use `nullptr`

ac101m: Ah, thought it might be something weird like this, just checking!

cpp/jvm-enclave-common/stubs/sys_poll.cpp
bonifazi-r3: I am still not 100% sure. ENOSYS exists and I think we can probably use. The JVM code does not seem care much about the `errno` value apart from some cases, just to throw the correct exception. There always seem to be a way to handle the returning result (i.e. `-1`), whatever the related `errno` is set to.

Still it is weird that there are some calls for which `ENOSYS` is a valid `errno` value and many others that do not mention it.
I have found this pdf about POSIX standard https://www.open-std.org/jtc1/sc22/open/n4217.pdf  (a bit old), and this suggest that `ENOSYS` has even been removed from some calls ass a valid `errno` value.
Even the Linux kernel does not seem to return `ENOSYS` quite often, see here: `https://github.com/torvalds/linux/blob/master/kernel/time/posix-stubs.c`, where it is used only in one case (not really even sure that that one is a Posix call).

But I have also found a similar case here https://support.kaspersky.com/help/kce/0.0/en-US/posix_uns_ifaces.htm where Kaspersky is doing exactly as you suggest, i.e. return -1 and `ENOSYS` as `errno`.

At the end of the day, what it matters is how the JVM reacts to that and I see that it is always reacting with an error as we want. I I feel that it is fine to do so.

cpp/jvm-enclave-common/stubs/dlfcn.cpp
bonifazi-r3: I don't see any errno code at all for them https://man7.org/linux/man-pages/man3/dlopen.3.html

PR #75, opened 2022-10-21
CON-1032: Deprecate EPID attestation

bonifazi-r3:
EPID attestation is not any more supported by modern Intel SGX-capable CPUs and therefore we are deprecating it.
This PR mark the related class as deprecated and updates the EPID related docs.


bonifazi-r3:
@shamsasari Is this ticket just these few changes or did you have more in mind?

conclave-host/src/main/kotlin/com/r3/conclave/host/AttestationParameters.kt
```diff
@@ -29,6 +29,10 @@ sealed class AttestationParameters {
      * This parameter is not used if the enclave is in simulation mode (as no attestation is done in simulation) and null
      * can be provided.
      */
+    @Deprecated(
+        "EPID attestation support is deprecated and it will be removed in the next release," +
```
shamsasari, line 33: not the next release, but in an upcoming release

docs/docs/faq.md
```diff
@@ -152,6 +152,9 @@ Intel will whitelist your signing key and the release mode then becomes availabl
     object to the clients. The Conclave client libraries embed the necessary certificates to verify Intel's signature over
     this data, and the integrity of the object is checked automatically when it's deserialized.
 
+!!! Important
+    Conclave is deprecating the EPID attestation protocol, it will be removed in the next release. You should use modern
+    hardware that supports the DCAP protocol.
 ### How are upgrades performed when patches are released?
```
shamsasari, line 158: Add a new line gap

docs/docs/faq.md
```diff
@@ -152,6 +152,9 @@ Intel will whitelist your signing key and the release mode then becomes availabl
     object to the clients. The Conclave client libraries embed the necessary certificates to verify Intel's signature over
     this data, and the integrity of the object is checked automatically when it's deserialized.
 
+!!! Important
```
shamsasari, line 155: The documentation shouldn't be introducing the concept of EPID anymore. Instead where applicable it should be changed to DCAP

scripts/ci_hardware_dcap_tests.sh
```diff
@@ -11,7 +11,6 @@ teardownAESM
 startAESMContainer
 
 # Run the DCAP attestation tests in Debug and Simulation modes using integration tests.
-# Note, the DCAP run doesn't test the samples as they can be run on the EPID machines which are cheaper.
```
shamsasari, line 14: This script, and any related to it, are no longer called and can be deleted. Since we moved to TC Cloud, all our hardware (i.e. debug mode) tests are done in DCAP since all the agents are in Azure.

shamsasari (changes requested):
The docs should be changed to refer only to DCAP. You will also need to update the release notes to state that EPID has been deprecated and will be removed in an upcoming release.

Also, you can remove the AESMD container as it was only needed for EPID. We don't need it for DCAP since we're doing in-process attestation.

docs/docs/faq.md
bonifazi-r3: Ok, so if we are deprecating we don't even want to give instructions for EPID, then? I.e. we want to "pretend" that it is not existing any more in the instructions, right? Why not demising then?

bonifazi-r3:
> The docs should be changed to refer only to DCAP. You will also need to update the release notes to state that EPID has been deprecated and will be removed in an upcoming release.
> 
> Also, you can remove the AESMD container as it was only needed for EPID. We don't need it for DCAP since we're doing in-process attestation.

@shamsasari  I am noticing there are some changes here to be merged https://github.com/R3Conclave/conclave-core-sdk/pull/76, which relate to docs. Can we get them merged or do we want to wait for John?

docs/docs/faq.md
shamsasari: Yes, because otherwise documentation implies customer support if they have issues, which is not something we want to be doing now.

shamsasari:
No, if you could have a quick look at the merge PR and approve. There aren't any changes in it expect for the release-notes, which John can take a look after he comes back.

docs/docs/non-cloud-deployment.md
```diff
@@ -84,32 +76,9 @@ A quick summary:
     * For Ubuntu 20 LTS: `echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu focal main' > /etc/apt/sources.list.d/intelsgx.list`
     * Add the Intel package signing key: `wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | apt-key add -`
     * Run `apt-get update`
-    * Run `apt-get install libssl-dev libcurl4-openssl-dev libprotobuf-dev libsgx-urts libsgx-launch libsgx-epid libsgx-quote-ex`
+    * Run `apt-get install libssl-dev libcurl4-openssl-dev libprotobuf-dev libsgx-urts libsgx-launch libsgx-quote-ex`
```
shamsasari, line 79: The integratlon-tests-build container has epid reference

bonifazi-r3: You mean in the `Dockerfile`, right?
Sorry, my IntelliJ does not pick up all the files when switching branches, removing that install package as well as well.

integration-tests/general/common-test/src/main/kotlin/com/r3/conclave/integrationtests/general/commontest/TestUtils.kt
```diff
@@ -12,19 +12,10 @@ object TestUtils {
...
-                AttestationParameters.EPID(spid, attestationKey)
-            }
-            else -> throw UnsupportedOperationException(
-                "SGX does not appear to be available on this machine. Check kernel drivers."
-            )
-        }
+        return if (Path.of("/dev/sgx_enclave").exists() || Path.of("/dev/sgx/enclave").exists())
+            AttestationParameters.DCAP()
+        else -> throw UnsupportedOperationException(
+        "SGX does not appear to be available on this machine. Check kernel drivers."
```
shamsasari, line 18: indent

integration-tests/general/common-test/src/main/kotlin/com/r3/conclave/integrationtests/general/commontest/TestUtils.kt
```diff
@@ -12,19 +12,10 @@ object TestUtils {
...
 
     private fun getHardwareAttestationParams(): AttestationParameters {
-        return when {
-            // EPID vs DCAP can be detected because the drivers are different and have different names.
-            Path.of("/dev/sgx_enclave").exists() || Path.of("/dev/sgx/enclave").exists() -> {
-                AttestationParameters.DCAP()
-            }
-            Path.of("/dev/isgx").exists() -> {
-                val spid = OpaqueBytes.parse(System.getProperty("conclave.spid"))
-                val attestationKey = checkNotNull(System.getProperty("conclave.attestation-key"))
```
shamsasari, line 22: Also remove these from build.gradle

integration-tests/general/common-test/src/main/kotlin/com/r3/conclave/integrationtests/general/commontest/TestUtils.kt
bonifazi-r3: This is actually how IntelliJ indent this automatically. I feel that it does not see the scope after `else`. I suggest to use brackets to make this easier
```
    private fun getHardwareAttestationParams(): AttestationParameters {
        return if (Path.of("/dev/sgx_enclave").exists() || Path.of("/dev/sgx/enclave").exists()) {
            AttestationParameters.DCAP()
        } else -> {
            throw UnsupportedOperationException(
                "SGX does not appear to be available on this machine. Check kernel drivers."
            )
        }
    }
```

shamsasari (approved):

shamsasari:
The integration tests are failing due to a compile error

PR #77, opened 2022-10-24
CON-1193 - conclave init - Added a logging framework to be used as default by the projects generated using the conclave-init tool

filipesoliveira:
The projects generated by `conclave-init` will now have a default logging framework. 
Updated io.spring.dependency-management to the latest version (1.1.0).
Added a check to ensure the host generates logs by default.

### DIFF
--- a/test-conclave-init.sh
+++ b/test-conclave-init.sh
@@ -39,7 +39,7 @@ sed -i "s/repositories {/repositories {\nmaven { url = '..\/repo' }/" settings.g
 echo Run Java project host and client
 
 # NOTE: this command is only run in mock mode, so that the signer can be provided below.
-./gradlew :host:run & _PID=$!
+./gradlew :host:bootRun & _PID=$!
 sleep 5
 ./gradlew :client:run \
   --args="'S:0000000000000000000000000000000000000000000000000000000000000000 PROD:1 SEC:INSECURE'"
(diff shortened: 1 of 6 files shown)


filipesoliveira:
_Can you look into updating test-conclave-init.sh to capture the output of the boot jar run to make sure it logged the missing bits._
Sorry, @shamsasari. I did not understand what you mean by "capture the output of the boot jar run to make sure it logged the missing bits."

shamsasari:
I mean, can we test that the logging is working? We test conclave init in the test-conclave-init.sh script. Is it possible to update it so that when the generated code is executed, we assert that the host process logged the lines which previously was not being logged. This should be possible by simply redirecting to grep (via a file or direct via a pipe).

test-conclave-init.sh
```diff
@@ -39,7 +39,7 @@ sed -i "s/repositories {/repositories {\nmaven { url = '..\/repo' }/" settings.g
 echo Run Java project host and client
 
 # NOTE: this command is only run in mock mode, so that the signer can be provided below.
-./gradlew :host:run & _PID=$!
+./gradlew :host:bootRun & _PID=$!
```
shamsasari, line 42: Do we need to update the docs?

filipesoliveira: Docs have been updated.

shamsasari:
Test failing for some reason.

filipesoliveira:
@shamsasari, all tests are now green.

shamsasari (approved):
LGTM

PR #78, opened 2022-10-26
[NOTICK] Call interface performance improvements

ac101m:
# Changes
- Add separate connections for each concurrent call "slot"


ac101m:
Rewrote the socket call interface to reduce complexity and contention.

@bonifazi-r3 Adding you on review as well as with these performance improvements, the socket interface may be around longer than asserted in the original PR.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
```diff
@@ -86,25 +49,17 @@ class SocketEnclaveHostInterface(
...
-
-                /** Read the maximum number of concurrent calls from the host. */
-                val maximumConcurrentCalls = fromHost.readInt()
-                callExecutor = Executors.newFixedThreadPool(maximumConcurrentCalls)
-
-                /** Start the message receive thread. */
-                receiveLoopThread.start()
+                /** Receive the maximum number of concurrent calls from the host and set up the call executor service. */
+                Socket(host, port).use { initialSocket ->
+                    maximumConcurrentCalls = initialSocket.getInputStream().readInt()
```
shamsasari, line 54: Can you wrap this inside `DataOutputStream`. Let's try to get rid of the read* methods we ended up creating

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -1133,7 +1133,7 @@ class EnclaveHost private constructor(
                     val pathMatcher = manifestSearchPattern.matcher(resource.path)
 
                     if (pathMatcher.matches()) {
-                        val enclaveClassName = pathMatcher.group(1).replace('/', '.')
+                        val enclaveClassName = pathMatcher.group(1).replace('/', '.').substringAfter("!.")
```
shamsasari, line 1136: Can you give an example of this. I feel this is a bug and we're just covering it up. Probably in the plugin.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -22,15 +22,12 @@ import java.util.concurrent.Semaphore
  *  - Handle the low-level details of the messaging protocol (socket with streamed ECalls and OCalls).
  */
 class SocketHostEnclaveInterface(port: Int = 0, private val maxConcurrentCalls: Int) : HostEnclaveInterface(), Closeable {
-    private val serverSocket = ServerSocket(port)
+    private val serverSocket = ServerSocket(port, 100)
```
shamsasari, line 25: magic number alert! Why are you having to increase the backlog value?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
```diff
@@ -122,41 +77,33 @@ class SocketEnclaveHostInterface(
...
-    private fun sendMessageToHost(message: SocketCallInterfaceMessage) {
-        val messageBytes = message.toByteArray()
-        synchronized(toHost) {
-            toHost.writeIntLengthPrefixBytes(messageBytes)
-            toHost.flush()
+    fun awaitTermination() {
+        var done = false
+        while(!done) {
+            try {
+                done = callExecutor.awaitTermination(10, TimeUnit.SECONDS)
```
shamsasari, line 88: Delegate the timeout parameter to the caller. Use a single `Duration` parameter. This will avoid all this messy termination logic

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
```diff
@@ -86,25 +49,17 @@ class SocketEnclaveHostInterface(
...
-                receiveLoopThread.start()
+                /** Receive the maximum number of concurrent calls from the host and set up the call executor service. */
+                Socket(host, port).use { initialSocket ->
+                    maximumConcurrentCalls = initialSocket.getInputStream().readInt()
+                    callExecutor = Executors.newFixedThreadPool(maximumConcurrentCalls)
+                }
+
+                /** Connect sockets and instantiate handler threads. */
+                for (i in 0 until maximumConcurrentCalls) {
+                    val socket = Socket(host, port).apply { tcpNoDelay = true }
```
shamsasari, line 60: Let's not prematurely create connections. Only do so when a new thread is detected.

shamsasari (changes requested):

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
ac101m: This was an accidental commit. There should not be any changes to EnclaveHost in this PR, will fix.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
ac101m: Another one that slipped through! Should just be set to the max number of concurrent calls. Will fix.

ac101m: Actually, as the socket creation process is synchronous, it can be left default. Will remove.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
ac101m: Done

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
shamsasari: Yes, it will require having a "control" connection to listen for new connection requests. If it's too much work then leave for now, but add a TODO, just in case this implementation gets traction.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
ac101m: The timeout isn't important here. This function is just supposed to wait until the interface terminates, regardless of how long that takes.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
ac101m: I guess this would reduce overhead in the case where there are only a few threads in the enclave at once. For now I've just added a TODO.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/SocketEnclaveHostInterface.kt
shamsasari: The implementation is not correct. You're not supposed to catch and ignore InterruptedException. Typically you rethrow into another exception, or more correctly you reinstate the interrupt flag for the current thread so that a higher caller can detect the interrupt and do something about it if they need to. Typically an interrupt is called to signal that the thread should stop executing.

If you simply delegate to `callExecutor.awaitTermination` with a user specified `Duration` then you avoid all this. Further, since this method is called from the main method, you can see quite clearly that there is no need to deal with interrupts, since it's the main thread executing. In fact a timeout duration doesn't even make sense in that context, which implies this method should have the timeout as an nullable parameter.

ac101m: Gotcha. 
I don't think there is an await termination without a timeout parameter (don't know why). I wouldn't have bothered with the loop in that case!

shamsasari: There isn't actually, my bad! Just pass in the timeout as a param and let the main method set a timeout, something like max int seconcds or soemthing.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -67,7 +67,9 @@ object GramineEntryPoint {
          */
         hostInterface.use {
             it.start()
-            it.awaitTermination()
+            it.awaitTermination(Long.MAX_VALUE)
         }
+
+        System.err.println("\nHost interface terminated!\n")
```
shamsasari, line 73: delete

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -24,13 +23,10 @@ import java.util.concurrent.Semaphore
 class SocketHostEnclaveInterface(port: Int = 0, private val maxConcurrentCalls: Int) : HostEnclaveInterface(), Closeable {
     private val serverSocket = ServerSocket(port)
```
shamsasari, line 24: The socket shouldn't be bound in the c'tor. In fact this socker is used and then closed in `start` you can move this there, and then have the `port` property as a lateinit which is set in `start`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
ac101m: Oops!

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
ac101m: We also need to pass the port to the enclave process, and we don't know what the port will be until it's bound, so there's that too.

shamsasari: Yes `start` should probably be split into two, `start` and `run`.

shamsasari: Sorry, I thought `start` blocked, but it doesn't. So no, I'm not sure why we would need a separate method. Looking at GramineEnclaveHandle, you would simply call `start` synchronously. I'm not sure why it needs to be done on a separate thread.

shamsasari: >We also need to pass the port to the enclave process, and we don't know what the port will be until it's bound, so there's that too.

I think this can be fixed if we say that the port is only defined after `start` is called. Then the can use pass it to the gramine process. This would require not starting the interface on a separate thread, which I don't think seems to be necessary.

ac101m: > Looking at GramineEnclaveHandle, you would simply call `start` synchronously. I'm not sure why it needs to be done on a separate thread.

Start blocks until the enclave connects, so it has to be called either asynchronously or after the enclave process has been started. In the current implementation (binding the port at construction), you're right, it probably doesn't need to be done in a thread like this!

ac101m: I've cleaned this up a bit, adding a separate `bindPort()` method and removing the unneeded thread in the enclave handle.
LMK if you think this is sufficient.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -41,6 +38,15 @@ class SocketHostEnclaveInterface(port: Int = 0, private val maxConcurrentCalls:
 
     private val callGuardSemaphore = Semaphore(0)
 
+    /**
+     * Set up the server socket, binding the specified port.
+     * If no specific port is requested, let the system allocate one.
+     */
+    fun bindPort(port: Int = 0): Int {
```
shamsasari, line 45: Technically you've introduced another state here.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
```diff
@@ -46,24 +46,22 @@ class GramineEnclaveHandle(
...
                 .inheritIO()
                 .directory(enclaveDirectory.toFile())
-                .command("gramine-direct", "java", "-cp", GRAMINE_ENCLAVE_JAR_NAME, "com.r3.conclave.enclave.internal.GramineEntryPoint", enclaveInterface.port.toString())
+                .command("gramine-direct", "java", "-cp", GRAMINE_ENCLAVE_JAR_NAME, "com.r3.conclave.enclave.internal.GramineEntryPoint", port.toString())
                 .start()
 
             /** Wait for the local call interface start process to complete. */
-            interfaceStartThread.join()     // wait for start process to finish
-            interfaceStartTask.get()        // throw if start failed
+            enclaveInterface.start()
```
shamsasari, line 64: This is much better. I think we can improve it even further by removing the synchronized blocks in `start` and `close`.

ac101m: Ehhh, not sure about this.
Maybe in start if we assume that no calls will arrive until `start()` has returned. Definitely not in `close()` though, as there's a race condition with the state check in `initiateOutgoingCall()`.
More generally I think the state manager could use a little work. With any state machine there is a process that takes it from one state to another. I think this could be better expressed by passing a lambda to the transition methods that contains the actions that need to accompany the state change. There's also possibly an error condition if the state transition fails. I also think all state transitions should all be atomic. It really doesn't make sense for a state machine to be operated on by multiple threads anyway.

shamsasari (approved):
Good job.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
ac101m: Close has to wait for currently running calls to finish. There is a race condition between changing the state to stopped and acquiring the concurrent call semaphore that could result in a thread (or threads) waiting forever in `initiateOutgoingCall()` otherwise.

ac101m:
@shamsasari Just discovered a potential issue with this (unrelated to your comments), moved back to draft temporarily.

PR #81, opened 2022-10-27
[CON-1194] Gramine: MRSIGNER and MRENCLAVE in simulation mode

ac101m:
# Changes
- The `MRENCLAVE` value in `GramineDirectEnclaveEnvironment` is now the hash of the enclave fat-jar.
- An "enclave metadata" properties file has been added to the gramine build process.
- Remove janky manifest-parsing process for getting the enclave thread count.
- Pass enclave thread count to the call interface using the new metadata file.
- The `MRSIGNER` value is now brought in using the new metadata file.


conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
```diff
@@ -3,7 +3,10 @@ package com.r3.conclave.enclave.internal
 import com.r3.conclave.common.EnclaveMode
 import com.r3.conclave.common.internal.*
 import com.r3.conclave.utilities.internal.digest
+import java.io.File
 import java.nio.ByteBuffer
+import java.nio.file.Path
```
bonifazi-r3, line 8: I think it is an unused import

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
```diff
@@ -31,14 +35,24 @@ class GramineDirectEnclaveEnvironment(
...
+
+        val buffer = ByteArray(4096)
+        var bytesRead: Int
+
+        val digest = MessageDigest.getInstance("SHA-256")
+
+        File(jarPath).inputStream().use {
+            bytesRead = it.read(buffer)
+            while (bytesRead >= 0) {
+                digest.update(buffer, 0, bytesRead)
```
bonifazi-r3, line 50: Dumb question, it looks fine to me, but wanted to double check my understanding.
Are we updating the digest many times (and I assume "updating" does not mean"replacing" or "overriding") by providing values from the buffer from 0 to 4096 (apart from the last read).
This is how it should work, right?
But is `4096` a conventional number for buffers or what? (i.e. maybe let's try to reduce the number of iterations?)

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/BuildUnsignedGramineEnclave.kt
```diff
@@ -44,10 +47,16 @@ open class BuildUnsignedGramineEnclave @Inject constructor(objects: ObjectFactor
     }
 
     private fun generateGramineDirectManifest(templateManifest: String) {
+        /**
+         * It's possible for Gramine to launch threads internally that conclave won't know about!
```
bonifazi-r3, line 51: `Conclave` upper case is nicer!
Also maybe `doesn't` instead of `won't`?

bonifazi-r3 (changes requested):
Looks pretty ok to me, minor questions

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
```diff
@@ -1,13 +1,13 @@
...
 import com.r3.conclave.common.EnclaveMode
-import java.io.BufferedReader
 import java.io.IOException
 import java.io.InputStreamReader
 import java.net.URL
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
+import java.util.*
 import java.util.concurrent.FutureTask
```
bonifazi-r3, line 11: Import

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
```diff
@@ -1,13 +1,13 @@
 package com.r3.conclave.host.internal
 
 import com.r3.conclave.common.EnclaveMode
-import java.io.BufferedReader
 import java.io.IOException
 import java.io.InputStreamReader
```
bonifazi-r3, line 5: Import here as well?

bonifazi-r3 (changes requested):
Two imports, will do another round later

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
ac101m: Yup, removed

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
ac101m: Yup, pretty much. "updating" here just means feeding more data into the hash function. I think this is pretty standard I think (see `java.security.MessageDigest`). As for 4096, there's no particular reason, I just didn't want to read the whole file into a byte array, and I like powers of two!

ac101m: I guess it could do to be a bit larger. What was the file block hashing size used by gramine again?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/BuildUnsignedGramineEnclave.kt
ac101m: Capitalisation, sure. But I think won't is the right word here. Definitely makes sense for it to be future tense at this point!

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -5,14 +5,46 @@ import com.r3.conclave.common.internal.CallHandler
...
+    private const val ENCLAVE_METADATA_FILE = "enclave-metadata.properties"
     private const val EXIT_ERR = -1
 
+    /** Enclave metadata variables */
+    private var isSimulation by Delegates.notNull<Boolean>()
+    private lateinit var signingKeyMeasurement: ByteArray
+
+    /** Load enclave metadata from the enclave properties file. */
+    private fun loadEnclaveMetadata() {
+        val properties = File(ENCLAVE_METADATA_FILE).reader().use {
```
shamsasari, line 26: Prefer `Path` over `File`

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -5,14 +5,46 @@ import com.r3.conclave.common.internal.CallHandler
...
+import kotlin.properties.Delegates
 import kotlin.system.exitProcess
 
 object GramineEntryPoint {
     private const val USAGE_STRING = "usage: GramineEntryPoint <port>"
+    private const val ENCLAVE_METADATA_FILE = "enclave-metadata.properties"
     private const val EXIT_ERR = -1
 
+    /** Enclave metadata variables */
+    private var isSimulation by Delegates.notNull<Boolean>()
```
shamsasari, line 21: What's Delegates?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -5,14 +5,46 @@ import com.r3.conclave.common.internal.CallHandler
...
+    private lateinit var signingKeyMeasurement: ByteArray
+
+    /** Load enclave metadata from the enclave properties file. */
```
shamsasari, line 24: Some explanation as to why we need the metadata file.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
bonifazi-r3: I guess this will "update" hundreds, if not thousands, times. Not really sure about "block hashing size".
But if we don't notice immediate performance issues, we can postpone the matter and leave this as it is

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -5,14 +5,46 @@ import com.r3.conclave.common.internal.CallHandler
...
+    private fun loadEnclaveMetadata() {
```
shamsasari, line 25: Minor point on code structure. I would expect the main method to be the first method in the file. Makes it easier to read as what the entry point is doing.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -134,40 +137,66 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+            )
+        }
+    }
+
+    private fun buildUnsignedGramineEnclaveTask(
+            target: Project,
+            type: BuildType,
+            conclaveExtension: ConclaveExtension
+    ): BuildUnsignedGramineEnclave {
+        val gramineBuildDir = baseDirectory.resolve("gramine").toString()
```
shamsasari, line 163: Can we move this out into a field or val, since the location is now used in multiple places, perhaps alongside `baseDirectory`.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.OutputFile
+import java.io.File
+import java.security.interfaces.RSAPublicKey
+import java.util.*
+import javax.inject.Inject
+
+/**
+ * This task generates a properties file containing metadata used by the host to launch a gramine enclave.
+ * It is also used in simulation mode to pass a mock mrsigner value to the enclave.
```
shamsasari, line 22: Just elaborate that no actual signing occurs in simulation mode.

shamsasari (changes requested):

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,100 @@
...
+        /**
+         * The modulus bytes are a big-endian representation.
+         * Here we do a quick sanity check to ensure that the MSB (which would contain the sign bit) really is zero.
+         */
+        check(modulusBytes[0] == 0.toByte())
+
+        /** Reverse the bytes (currently big endian, need little endian), then compute the measurement. */
+        return digest("SHA-256") {
+            modulusBytes.reverse()
+            update(modulusBytes, 0, 384)    // Ignore the 385th byte.
```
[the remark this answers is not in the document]
ac101m: Previous comments in this method. Will make this a little clearer.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
ac101m: Alternative to `lateinit` that works for primitive types. Was actually suggested to me by the IDE!

bonifazi-r3: I agree, it's a bit odd. I would not trust IntelliJ sometimes.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
```diff
@@ -31,14 +34,24 @@ class GramineDirectEnclaveEnvironment(
...
+                bytesRead = it.read(buffer)
+            }
```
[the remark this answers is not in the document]
bonifazi-r3: Looks like the way Alex has implemented is common. Not sure if there is any other way, but it looks fine.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
ac101m: I really don't think it's a problem, this only happens once on startup. I've bumped it to 64k for now.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,107 @@
...
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.OutputFile
+import java.io.File
+import java.security.interfaces.RSAPublicKey
+import java.util.*
+import javax.inject.Inject
+
+/**
+ * In native mode, enclave metadata is bundled into the enclave .so file by the intel SGX SDK. The Intel SDK then uses
```
bonifazi-r3, line 21: Intel uppercase is nicer

bonifazi-r3 (changes requested):
Last minor comment from my side, after addressing Sham's concerns I think this is fine

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
ac101m: Removed, just used nullable values instead for both of them

bonifazi-r3 (commented):
@shamsasari It is fine for me, I'll leave it to you for a final approve

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -38,29 +71,71 @@ object GramineEntryPoint {
         return port
     }
 
+    /**
+     * Load enclave metadata properties file.
+     * This file contains properties required to load the enclave, such as the signing key measurement, threading level,
+     * and whether to use gramine-sgx or gramine-direct.
```
shamsasari, line 77: The later point of gramine-sgx vs gramine-direct is not relevant here and my confuse someone later looking at this code.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -38,29 +71,71 @@ object GramineEntryPoint {
...
+     * For more information, see [com.r3.conclave.plugin.enclave.gradle.gramine.GenerateGramineEnclaveMetadata].
+     */
+    private fun loadEnclaveMetadata() {
+        val properties = Paths.get(ENCLAVE_METADATA_FILE).toFile().reader().use { reader ->
```
shamsasari, line 81: `toFile()` not necessary

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -26,7 +26,7 @@ fs.mounts = [
 
 sgx.enclave_size = "4G"
 
-sgx.thread_num = 64
+sgx.thread_num = {{ maxThreads }}
```
shamsasari, line 29: `max_threads` for consistency

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        outputGramineEnclaveMetadata.asFile.get().writer().use { writer ->
+            for (entry in properties.entries) {
+                writer.write("${entry.key}=${entry.value}\n")
+            }
+        }
+    }
+
+    /** Compute the mrsigner value from a provided .pem file containing a 3072 bit RSA key. */
+    private fun computeSigningKeyMeasurement(keyFile: File): ByteArray {
```
shamsasari, line 85: No need to specify a parameter when this is an instance method and has access to the input file.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    @get:Input
+    val buildType: Property<BuildType> = objects.property(BuildType::class.java)
+    @get:Input
+    val maxThreads: Property<Int> = objects.property(Int::class.java)
+    @get:InputFile
+    val signingKey: RegularFileProperty = objects.fileProperty()
+    @Optional
+    @get:Input
+    val pythonEnclaveFile: Property<String> = objects.property(String::class.java)
```
shamsasari, line 45: This is just set to the "enclave.py" constant so there's no benefit to having this. Instead this should be a boolean.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * TODO: Investigate using environment variables in the Manifest to pass this metadata instead.
+ */
+open class GenerateGramineEnclaveMetadata @Inject constructor(objects: ObjectFactory) : ConclaveTask() {
+    companion object {
+        const val METADATA_FILE_NAME = "enclave-metadata.properties"
+    }
+
+    @get:Input
+    val buildType: Property<BuildType> = objects.property(BuildType::class.java)
```
shamsasari, line 38: Inject this in. The way the plugin is designed is such each mode operates independently of each other, meaning you can create all four modes of the enclave without overwriting each other. Which all means this isn't an input property to tasks (you can see that if the other tasks).

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,117 @@
...
+    @get:Input
+    val maxThreads: Property<Int> = objects.property(Int::class.java)
+    @get:InputFile
+    val signingKey: RegularFileProperty = objects.fileProperty()
+    @Optional
+    @get:Input
+    val pythonEnclaveFile: Property<String> = objects.property(String::class.java)
+
+    @get:OutputFile
+    val outputGramineEnclaveMetadata: RegularFileProperty = objects.fileProperty()
```
shamsasari, line 48: Let's turn away from this pattern of "input"/"output" prefixes. I don't think it's necessary if the name of the property is clear and with docs if necessary.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    override fun action() {
+        val isSimulation = buildType.get() == BuildType.Simulation
+
+        /**
+         * We don't use the java Properties class here for two reasons:
+         * - It doesn't guarantee consistent order.
+         * - It prints the date and time at the top of the file.
+         */
```
shamsasari, line 57: This is the second time we're doing this. Add a helper function in Utils which takes in a lamdba and a `Path` and will write out the contents. Something like

```
writeStableProperties(path) {
  put("key", "value)
....
}
```

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineEnclaveMetadata.kt
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        outputGramineEnclaveMetadata.asFile.get().writer().use { writer ->
+            for (entry in properties.entries) {
+                writer.write("${entry.key}=${entry.value}\n")
+            }
+        }
+    }
+
+    /** Compute the mrsigner value from a provided .pem file containing a 3072 bit RSA key. */
```
shamsasari, line 84: Is there documentation you can link to help the next person who might look at this code?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -180,8 +182,32 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+    }
+
+    private fun createGenerateGramineManifestTask(
+            target: Project,
+            type: BuildType,
+            conclaveExtension: ConclaveExtension
+    ): GenerateGramineManifest {
         return target.createTask("generateGramineManifest$type") { task ->
+            task.maxThreads.set(conclaveExtension.maxThreads)
             task.manifestFile.set((baseDirectory / "gramine" / PluginUtils.GRAMINE_MANIFEST).toFile())
```
shamsasari, line 211: This can use your new `gramineBuildDirectory`

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
```diff
@@ -30,12 +30,12 @@ class GramineEnclaveHandle(
...
 
-        /**
-         * Not all threads in the enclave are necessarily the result of conclave calls.
-         * To minimise the likelihood of deadlocks, we don't allow conclave to use all available threads.
-         */
-        val maxConcurrentCalls = getEnclaveThreadCountFromManifest() / 2
-        check(maxConcurrentCalls > 0)
+        /** Load gramine enclave metadata. */
+        val metadataFile = workingDirectory / GRAMINE_ENCLAVE_METADATA
+        val maxConcurrentCalls = metadataFile.toFile().inputStream().use { inputStream ->
```
shamsasari, line 35: `toFile` not needed

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/GramineEnclaveHandle.kt
```diff
@@ -30,12 +30,12 @@ class GramineEnclaveHandle(
...
+            val metadataProperties = Properties().apply { load(inputStream) }
```
shamsasari, line 36: Have a look at the docs for `Properties.reader`. Is it safe to use this considering we're writing the strings without any of the weird escaping that Properties does? We might need to simply roll our own "properties" reading code.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -38,29 +71,71 @@ object GramineEntryPoint {
...
+            Properties().apply { load(reader) }
+        }
+
+        val isSimulationString = checkNotNull(properties["isSimulation"]).toString()
+        check(isSimulationString == "true" || isSimulationString == "false")
+        isSimulation = isSimulationString == "true"
```
shamsasari, line 87: Let's not use strings in place of booleans: `java.lang.Boolean.parseBoolean`

shamsasari (changes requested):
Looking good, a few more changes needed.

I also noticed that this PR has fixed all the failing KDS and mail integration tests. They were failing due to the key policy not matching on the MRSIGNER, so that's good.

shamsasari:
I've figured out what's been bugging me... Why is there a separate properties file when the existing enclave properties can also be used? The handle can read the enclave.properties by looking into the enclave fat jar, and because it's in the fat jar, the enclave entry point can read it by simply loading it as a resource, which is what we already do elsewhere. As a nice benefit it becomes one less file for gramine to deal with.

ac101m:
> I've figured out what's been bugging me... Why is there a separate properties file when the existing enclave properties can also be used? The handle can read the enclave.properties by looking into the enclave fat jar, and because it's in the fat jar, the enclave entry point can read it by simply loading it as a resource, which is what we already do elsewhere. As a nice benefit it becomes one less file for gramine to deal with.

Yeah, I've had similar thoughts. The reason we can't use the internal properties file for this (at the moment at least) is because some of this data needs to be available _before_ the enclave can be loaded. I have some vague ideas as to how we can refactor this so that there is only one properties file, but I haven't thought it through fully yet. Pretty sure it a non-trivial refactor though. Might make sense to roll it into [CON-1182](https://r3-cev.atlassian.net/browse/CON-1182).

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineManifest.kt
```diff
@@ -37,13 +42,20 @@ open class GenerateGramineManifest @Inject constructor(objects: ObjectFactory) :
             .single { it.startsWith("Location: ") }
             .substringAfter("Location: ")
 
+        /**
+         * It's possible for a Gramine enclave to launch threads internally that Conclave won't know about!
+         * Because of this, we need to add some safety margin.
+         */
+        val gramineMaxThreads = maxThreads.get() * 2
```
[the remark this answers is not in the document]
ac101m: Changed to just +8

bonifazi-r3:
There's a lot of overlapping with my PR, it will be a bloodbath :-)
One point: I see that you pass a flag to set the fact that the enclave is in simulation or not.
This is fine, but just note that in DEBUG and RELEASE mode we will have a look at another manifest file, the `java.manifest.sgx.
I.e. the presence of the manifest file `java.manifest` already tells us that we are in simulation mode. I used that thing to distinguish between simulation vs non-simulation mode in my PR

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -3,33 +3,32 @@ package com.r3.conclave.enclave.internal
...
-    private const val ENCLAVE_METADATA_FILE = "enclave-metadata.properties"
     private const val EXIT_ERR = -1
 
-    /** Enclave metadata variables */
-    private var isSimulation: Boolean? = null
-    private var signingKeyMeasurement: ByteArray? = null
-    private var pythonEnclaveScript: Path? = null
+    /** Enclave metadata, retrieved from the manifest. */
+    private val isSimulation = parseBoolean(System.getenv("CONCLAVE_IS_SIMULATION_ENCLAVE")!!)
+    private val isPythonEnclave = parseBoolean(System.getenv("CONCLAVE_IS_PYTHON_ENCLAVE")!!)
```
shamsasari, line 24: `Boolean.getBoolean` - very badly named but it does what you're doing here.

ac101m:
> There's a lot of overlapping with my PR, it will be a bloodbath :-) One point: I see that you pass a flag to set the fact that the enclave is in simulation or not. This is fine, but just note that in DEBUG and RELEASE mode we will have a look at another manifest file, the `java.manifest.sgx. I.e. the presence of the manifest file `java.manifest` already tells us that we are in simulation mode. I used that thing to distinguish between simulation vs non-simulation mode in my PR

This is, if you ask me, another reason why we should have a universal metadata file that exists in all modes, determining things by looking at the presence or absence of files is just messy if you ask me...

As for the bloodbath, I blame shams for not reviewing while I was on PTO :laughing:

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -3,33 +3,32 @@ package com.r3.conclave.enclave.internal
...
+    private val signingKeyMeasurement = System.getenv("CONCLAVE_SIMULATION_MRSIGNER")?.let { Base64.getDecoder().decode(it) }
+    private val conclaveMaxThreads = System.getenv("CONCLAVE_MAX_THREADS")!!.toInt()
```
shamsasari, line 26: Similarly, `Integer.getInteger`

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -64,9 +66,9 @@ class SocketHostEnclaveInterface(private val maxConcurrentCalls: Int) : HostEncl
                 /** Wait for IO sockets, then close the server socket. */
                 serverSocket.use {
 
-                    /** Send the maximum number of concurrent calls to the enclave. */
+                    /** Receive the maximum number of concurrent calls from the enclave. */
```
shamsasari, line 69: Much better.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineManifest.kt
```diff
@@ -15,8 +23,14 @@ open class GenerateGramineManifest @Inject constructor(objects: ObjectFactory) :
         const val MANIFEST_TEMPLATE = "java.manifest.template"
     }
 
+    @get:Input
+    val buildType: Property<BuildType> = objects.property(BuildType::class.java)
```
shamsasari, line 27: Inject

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -14,6 +14,20 @@ loader.env.LD_LIBRARY_PATH = "/lib:{{ arch_libdir }}:/usr{{ arch_libdir }}:/usr/
...
+# Is this a simulation enclave?
+loader.env.CONCLAVE_IS_SIMULATION_ENCLAVE = "{{ is_simulation_enclave }}"
+
+# In simulation mode, we need to provide the signing key measurement
+{% if is_simulation_enclave == "true" %}
+  loader.env.CONCLAVE_SIMULATION_MRSIGNER = "{{ simulation_mrsigner }}"
+{% endif %}
+
+# Number of call interface worker threads, this needs to be less than gramine_max_threads
+loader.env.CONCLAVE_MAX_THREADS = "{{ conclave_max_threads }}"
```
shamsasari, line 29: Maybe then call this worker thread count?

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
```diff
@@ -7,5 +7,6 @@ object PluginUtils {
     const val ENCLAVE_PROPERTIES = "enclave.properties"
     const val GRAMINE_ENCLAVE_JAR = "enclave.jar"
     const val GRAMINE_MANIFEST = "java.manifest"
+    const val GRAMINE_ENCLAVE_METADATA = "enclave-metadata.properties"
```
shamsasari, line 10: You can delete this now?

shamsasari (changes requested):

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineManifest.kt
ac101m: What do you mean by "inject"?

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
ac101m: Yup!

shamsasari:
>java.manifest.sgx. I.e. the presence of the manifest file java.manifest` already tells us that we are in simulation mode. I used that thing to distinguish between simulation vs non-simulation mode in my PR 

@bonifazi-r3 in your PR then perhaps you replace the simulation env with this check

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineManifest.kt
shamsasari: Use dependency injection. Have a look at the other tasks which take in the build type .

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineManifest.kt
```diff
@@ -37,18 +56,67 @@ open class GenerateGramineManifest @Inject constructor(objects: ObjectFactory) :
...
         commandLine(
             listOf(
                 "gramine-manifest",
                 "-Djava_home=${System.getProperty("java.home")}",
                 "-Darch_libdir=/lib/$architecture",
                 "-Dld_preload=$ldPreload",
                 "-Dpython_packages_path=$pythonPackagesPath",
+                "-Dis_python_enclave=${pythonEnclave.get()}",
+                "-Dis_simulation_enclave=${buildType.get() == BuildType.Simulation}",
+                "-Dsimulation_mrsigner=${String(Base64.getEncoder().encode(computeSigningKeyMeasurement()))}",
```
shamsasari, line 75: Use hex. It's how it's represented in the rest of the code base and thus make things easier to debug.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
ac101m: Don't these get system properties? We need to get environment variables.

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
ac101m: Done

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineManifest.kt
ac101m: Done

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineManifest.kt
```diff
@@ -13,6 +23,15 @@ open class GenerateGramineManifest @Inject constructor(objects: ObjectFactory) :
         const val MANIFEST_TEMPLATE = "java.manifest.template"
     }
 
+    @get:Input
+    val buildType: Property<BuildType> = objects.property(BuildType::class.java)
+    @get:Input
+    val maxThreads: Property<Int> = objects.property(Int::class.java)
+    @get:InputFile
```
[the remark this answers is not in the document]
ac101m: Added spaces.

Was grouping these by input/output.

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -14,6 +14,20 @@ loader.env.LD_LIBRARY_PATH = "/lib:{{ arch_libdir }}:/usr{{ arch_libdir }}:/usr/
 loader.env.LD_PRELOAD = "{{ ld_preload }}"
 loader.env.PYTHONPATH = "{{ python_packages_path }}"
 
+# Signal to the enclave entry point that we
```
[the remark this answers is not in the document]
ac101m: Yup, will fix

bonifazi-r3 (approved):

PR #84, opened 2022-11-01
CON-1163: Integrate Gramine SGX mode

bonifazi-r3:


bonifazi-r3:
@shamsasari Feel free to provide feedback. Your python enclave is failing, will have a look when I have time.

shamsasari (commented):
Had a quick scan of the changes. Main question is why is there a separate bundle and handle for gramine-sgx? Surely the zip structure for both is the same, and the only difference will be in the contents of the manifest? Otherwise there seems to be a lot of duplication.

Also, please create Gramine debug job on TC so that we can start tracking the tests for that as well.

bonifazi-r3:
> Had a quick scan of the changes. Main question is why is there a separate bundle and handle for gramine-sgx? Surely the zip structure for both is the same, and the only difference will be in the contents of the manifest? Otherwise there seems to be a lot of duplication.
> 
> Also, please create Gramine debug job on TC so that we can start tracking the tests for that as well.

Yes, that is a valid question and I don't think that there is a "perfect" way of doing that.

Just some notes:
- in SGX the content of the zip is different. In SGX we need to have the `.sig`, the `.token` and the `manifest.sgx`. We don't need/use the `.manifest` file when running the code, we only need that in the build phase.
- In the SGX build, the `enclave.jar` file also needs to be present/available in the phase of building, to generate the hash. 
- in SGX case we also have a different executable at the run phase: `gramine-sgx`.

There is room for some generalization, for example in retrieving the number of threads, but this also depends on @ac101m changes as well. I know that he was working on a different way of passing the number of threads, for example, but we can also read this directly from the manifest using a TOML library and we can have an abstract class or a helper for this

There is the possibility of having a single "Handle" file, but this would contain quite a lot of conditional code as well (recognizing the SGX vs DIRECT depending on the content of the bundle zip, for example)
 In the long term, I we might want to interact with some sgx related only fields of the manifest, this is the main reason I felt that SGX deserved its own flow somehow.

bonifazi-r3:
@shamsasari Feel free to have another look. This is currently working in debug mode, where I appear to be able to run the host in SGX. Not sure what to do about release mode: what is the `prepareForRelease` task for? I.e. I am not sure how to use the `externalKey`, how that mechanism works (if we use a privateKey it works in release mode as well)

bonifazi-r3:
@shamsasari This is probably not going in in this way, as there is a lot overlap and decisions to me made after @ac101m 's changes, but your Python Enclave Test seems to be working with this, let me know otherwise.
We will need to check memory parameters and found a nicer way to include Python libs in the trusted section

```
Starting> PythonEnclaveTest.receive_enclave_mail() (7.3)
Gramine is starting. Parsing TOML manifest file, this may take some time...
Detected a huge manifest, preallocating 64MB of internal memory.
-----------------------------------------------------------------------------------------------------------------------
Gramine detected the following insecure configurations:

  - sgx.debug = true                           (this is a debug enclave)
  - loader.insecure__use_cmdline_argv = true   (forwarding command-line args from untrusted host to the app)

Gramine will continue application execution, but this configuration must not be used in production!
-----------------------------------------------------------------------------------------------------------------------

Ending> PythonEnclaveTest.receive_enclave_mail() (7.3)
Starting> PythonEnclaveTest.receive_from_untrusted_host() (7.4)
Gramine is starting. Parsing TOML manifest file, this may take some time...
Detected a huge manifest, preallocating 64MB of internal memory.
-----------------------------------------------------------------------------------------------------------------------
Gramine detected the following insecure configurations:

  - sgx.debug = true                           (this is a debug enclave)
  - loader.insecure__use_cmdline_argv = true   (forwarding command-line args from untrusted host to the app)

Gramine will continue application execution, but this configuration must not be used in production!
-----------------------------------------------------------------------------------------------------------------------

Ending> PythonEnclaveTest.receive_from_untrusted_host() (7.4)

Gradle Test Executor 7 finished executing tests.

> Task :python-enclave:host:test
Finished generating test XML results (0.0 secs) into: /home/marco/conclave-core-sdk/integration-tests/python-enclave/host/build/test-results/test
Generating HTML test report...
Finished generating test html results (0.001 secs) into: /home/marco/conclave-core-sdk/integration-tests/python-enclave/host/build/reports/tests/test
Watching 93 directories to track changes
Watching 98 directories to track changes
Watching 99 directories to track changes
:python-enclave:host:test (Thread[Daemon worker Thread 6,5,main]) completed. Took 1 mins 8.046 secs.
w: Detected multiple Kotlin daemon sessions at build/kotlin/sessions

BUILD SUCCESSFUL in 1m 20s
39 actionable tasks: 27 executed, 12 up-to-date
marco@marcovm:~/conclave-core-sdk/integration-tests$ find -name *sgx
```

ac101m (commented):
Would prefer to see separate classes for sgx/direct enclave types as the code for these will be very different later once attestation is implemented. Otherwise, looks good to me.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -6,13 +6,14 @@ import com.r3.conclave.utilities.internal.digest
 import java.nio.ByteBuffer
 
 /**
- * This class is the enclave environment intended for use with gramine-direct.
+ * This class is the enclave environment intended for use with gramine-direct or gramine-sgx.
  * TODO: Currently the behaviour here is lifted directly from [MockEnclaveEnvironment].
  *       This will eventually need to behave the same way as native simulation mode does.
  */
-class GramineDirectEnclaveEnvironment(
+class GramineEnclaveEnvironment(
```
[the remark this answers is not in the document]
bonifazi-r3: @shamsasari This change was part of our discussion, but I think that Alex is right. We should use one bundle mechanism, but one environment for Simulation mode and another for Release/Debug. What do you think?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -41,20 +45,21 @@ object GramineEntryPoint {
     private fun initialiseEnclave(
         enclaveClassName: String,
         hostInterface: SocketEnclaveHostInterface,
-        pythonScript: Path?
+        pythonScript: Path?,
```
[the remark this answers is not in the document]
bonifazi-r3: Agreed, a lot of this code will change after your PR goes in.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -41,20 +45,21 @@ object GramineEntryPoint {
...
+        enclaveMode: EnclaveMode
     ) {
         val enclaveClass = Class.forName(enclaveClassName)
         val enclave = enclaveClass.asSubclass(Enclave::class.java)
-                .getDeclaredConstructor()
-                .apply { isAccessible = true }
-                .newInstance()
+            .getDeclaredConstructor()
+            .apply { isAccessible = true }
+            .newInstance()
```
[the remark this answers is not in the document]
ac101m: It's a Java thing I think. The purpose of it is to make it easy to distinguish between normal indentation and wrapping that has been done to preserve horizontal space. See [here](https://www.oracle.com/java/technologies/javase/codeconventions-indentation.html), specifically the bullet point "If the above rules lead to confusing code or to code that's squished up against the right margin, just indent 8 spaces instead". I think most editors just default to eight spaces all the time these days. Additionally, you'll find if you put a newline in the middle of something like a parameter list, you'll notice that intellij will automatically double indent rather than single.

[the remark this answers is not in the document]
bonifazi-r3: That is my point: my editor does not default to that :-)! I see what you mean and I would be fine with that, but I think we should establish a common convention, with convention file included. Plus a TC job complaining about indentation rules.
Also noticing that Shams in his Plugin cleanup is converting from 8 to 4 spaces!
https://github.com/R3Conclave/conclave-core-sdk/pull/90/files#diff-41d01a82255cd13f751ee17782380ef3a1b5c17095358ac70ab00acb6fc557e6R26

bonifazi-r3:
The python enclave test seems to work now @shamsasari . There is a "number of thread" thing that we need to improve with @ac101m . I reopen the PR for review.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -9,10 +9,11 @@ import java.security.MessageDigest
...
  * This class is the enclave environment intended for use with gramine-direct.
  */
-class GramineDirectEnclaveEnvironment(
-        enclaveClass: Class<*>,
-        override val hostInterface: SocketEnclaveHostInterface,
-        private val mrsigner: ByteArray
+class GramineEnclaveEnvironment(
+    enclaveClass: Class<*>,
+    override val hostInterface: SocketEnclaveHostInterface,
+    private val simulationMrSigner: ByteArray,
```
shamsasari, line 15: Can you use `SecureHash` to make this a bit more typesafe. It has a `buffer()` method which you can use when setting the `mrsigner` field.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -9,10 +9,11 @@ import java.security.MessageDigest
...
+    override val enclaveMode: EnclaveMode
```
shamsasari, line 16: Add a `requires` check that this isn't mock.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
shamsasari, line 15: Minor: "mrsigner" is one word so it will be `*Mrsigner`

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -20,10 +23,12 @@ object GramineEntryPoint {
...
 
     /** Enclave metadata, retrieved from the manifest. */
-    private val isSimulation = parseBoolean(System.getenv("CONCLAVE_IS_SIMULATION_ENCLAVE")!!)
+    private val enclaveMode = EnclaveMode.valueOf(System.getenv("CONCLAVE_ENCLAVE_MODE")!!.uppercase())
     private val isPythonEnclave = parseBoolean(System.getenv("CONCLAVE_IS_PYTHON_ENCLAVE")!!)
-    private val signingKeyMeasurement = System.getenv("CONCLAVE_SIMULATION_MRSIGNER")?.let { parseHex(it) }
+    private val isSgxDebug = parseBoolean(System.getenv("SGX_DEBUG"))
+    private val simulationMrSigner = System.getenv("CONCLAVE_SIMULATION_MRSIGNER")?.let { parseHex(it) }
     private val conclaveWorkerThreads = System.getenv("CONCLAVE_ENCLAVE_WORKER_THREADS")!!.toInt()
+    private val commonErrorMessage = "Manifest not present when initialising the enclave in $enclaveMode mode"
```
shamsasari, line 31: Convert this into a function, otherwise this has unnecessarily created and stored a string

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -85,19 +90,33 @@ object GramineEntryPoint {
...
+                require(isSgxDebug) { "Gramine SGX debug flag not enabled in $enclaveMode mode" }
+            }
+
+            EnclaveMode.RELEASE -> {
+                require(File(GRAMINE_SGX_MANIFEST).exists()) { "Gramine SGX $commonErrorMessage" }
+                require(!isSgxDebug) { "Gramine SGX debug flag not disabled in $enclaveMode mode" }
+            }
+
+            EnclaveMode.MOCK -> {
+                throw IllegalArgumentException("Developer error: MOCK mode not supported in Gramine")
```
shamsasari, line 118: If we get here then I think it's a bug in the SDK. I don't think the user can recreate this, so remove "developer error"

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
```diff
@@ -20,10 +23,12 @@ object GramineEntryPoint {
     private const val EXIT_ERR = -1
 
     /** Enclave metadata, retrieved from the manifest. */
-    private val isSimulation = parseBoolean(System.getenv("CONCLAVE_IS_SIMULATION_ENCLAVE")!!)
+    private val enclaveMode = EnclaveMode.valueOf(System.getenv("CONCLAVE_ENCLAVE_MODE")!!.uppercase())
     private val isPythonEnclave = parseBoolean(System.getenv("CONCLAVE_IS_PYTHON_ENCLAVE")!!)
-    private val signingKeyMeasurement = System.getenv("CONCLAVE_SIMULATION_MRSIGNER")?.let { parseHex(it) }
+    private val isSgxDebug = parseBoolean(System.getenv("SGX_DEBUG"))
```
shamsasari, line 28: Isn't this redundant in the presence of `enclaveMode`?

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -414,8 +415,9 @@ class EnclaveHost private constructor(
         hostStateManager.checkStateIsNot<Closed> { "The host has been closed." }
 
         // This can throw IllegalArgumentException which we don't want wrapped in a EnclaveLoadException.
-        attestationService = AttestationServiceFactory.getService(enclaveMode, attestationParameters)
-        quotingService = EnclaveQuoteServiceFactory.getService(attestationParameters?.takeIf { enclaveMode.isHardware })
+        attestationService = AttestationServiceFactory.getService(enclaveMode, attestationParameters, enclaveHandle)
+        quotingService =
```
shamsasari, line 419: Add a TODO that this will need to fixed properly when Gramine attestation is introduced.

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -465,14 +467,19 @@ class EnclaveHost private constructor(
 
     private fun prepareFileSystemHandler(enclaveFileSystemFile: Path?): FileSystemHandler? {
         // TODO: Improve the enclaveHandle !is GramineEnclaveHandle condition
```
shamsasari, line 469: Delete TODO

conclave-host/src/main/kotlin/com/r3/conclave/host/EnclaveHost.kt
```diff
@@ -465,14 +467,19 @@ class EnclaveHost private constructor(
...
+        return if (isFileSystemSupported()) {
             val fileSystemFilePaths = if (enclaveFileSystemFile != null) listOf(enclaveFileSystemFile) else emptyList()
             FileSystemHandler(fileSystemFilePaths, enclaveMode)
         } else {
             null
         }
     }
 
+    private fun isFileSystemSupported(): Boolean {
+        //  This filesystem implementation is only supported in SIMULATION, DEBUG, RELEASE mode in Graal VM mode only
```
shamsasari, line 479: This is a TODO as well (for adding fs support in Gramine).

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/AttestationServiceFactory.kt
```diff
@@ -10,13 +12,20 @@ object AttestationServiceFactory {
...
-            EnclaveMode.SIMULATION -> MockAttestationService(isSimulation = true)
-            EnclaveMode.MOCK -> MockAttestationService(isSimulation = false)
-        }
+    fun getService(
+        enclaveMode: EnclaveMode,
+        attestationParameters: AttestationParameters?,
+        enclaveHandle: EnclaveHandle
+    ): AttestationService {
+        // TODO: Fix attestation in Gramine: https://r3-cev.atlassian.net/browse/CON-1178
+        return if (enclaveHandle is GramineEnclaveHandle) MockAttestationService(isSimulation = true)
```
shamsasari, line 21: I'd prefer this logic to be in `EnclaveHost.start`. It's not obvious to me that this `getService` will need to be affected by the Gramine attestation work. Or at least I'm not sure passing in the handle is the correct approach. Better to deal with this in that ticket.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -18,15 +24,28 @@ class GramineEnclaveHandle(
     override val enclaveClassName: String,
     private val zipFileUrl: URL
 ) : EnclaveHandle {
-    private lateinit var processGramineDirect: Process
+
+    companion object {
+        private val ENCLAVE_MODE_EXECUTABLES = mapOf(
```
shamsasari, line 29: This is better served as a function with a `when` statement. Saves on keeping the Map around.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -83,7 +106,23 @@ class GramineEnclaveHandle(
         }
     }
 
-    private fun unzipIntoWorkingDir() {
+    private fun getManifestFromUnzippedBundle(): Path {
+        check(enclaveMode != EnclaveMode.MOCK)
```
shamsasari, line 110: This is already checked for in the c'tor.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -83,7 +106,23 @@ class GramineEnclaveHandle(
...
+
+        return if (enclaveMode == EnclaveMode.SIMULATION) {
+            require((workingDirectory / GRAMINE_MANIFEST).exists()) { "Missing Gramine manifest" }
+            require((workingDirectory / GRAMINE_ENCLAVE_JAR).exists()) { "Missing enclave jar" }
```
shamsasari, line 114: We can avoid the duplicate check below

integration-tests/python-enclave/signing/sample_private_key.pem
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
shamsasari, line 1: Any particular reason why we need a fixed signing key? If we can avoid it then it's one less file

integration-tests/python-enclave/enclave/build.gradle
```diff
@@ -5,4 +5,15 @@ plugins {
 conclave {
     productID = 100
     revocationLevel = 2
+    maxThreads = 32
```
shamsasari, line 8: Why this many? Is this a requirement from Python?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -185,49 +192,136 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+    ): GenerateGramineDirectManifest {
         return target.createTask("generateGramineManifest$type", type) { task ->
+            task.inputSGXDebugFlag.set(
+                when (type) {
+                    BuildType.Simulation -> true
+                    BuildType.Debug -> true
+                    BuildType.Release -> false
+                    else -> true
+                }
+            )
```
shamsasari, line 204: This can be derived inside the task logic based on the `type` that's passed in.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
```diff
@@ -64,17 +69,19 @@ open class GenerateGramineManifest @Inject constructor(
          * Because of this, we need to add some safety margin.
          */
         val enclaveWorkerThreadCount = maxThreads.get()
-        val gramineMaxThreads = enclaveWorkerThreadCount + 8
+        //  TODO: Create a new parameters to distinguish between proper Enclave threads and threads used in the Enclave/Host Interface
+        val gramineMaxThreads = enclaveWorkerThreadCount * 2
```
shamsasari, line 73: Why does it need to be doubled?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
```diff
@@ -0,0 +1,89 @@
...
+    val inputPythonSourcePath: RegularFileProperty = objects.fileProperty()
+
+    @get:OutputFile
+    val outputSGXManifest: RegularFileProperty = objects.fileProperty()
+
+    @get:OutputFile
+    val outputToken: RegularFileProperty = objects.fileProperty()
+
+    @get:OutputFile
+    val outputSig: RegularFileProperty = objects.fileProperty()
```
shamsasari, line 41: Let's not follow the pattern of "input"/"output" prefixes. It's unnecessary as the context of the task should make it clear. This is not something common in Gradle plugins I've seen, just an odd decission that was made long ago in this repo. I'm in the process of removing these.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
```diff
@@ -0,0 +1,89 @@
...
+        if (inputPythonSourcePath.isPresent) {
+            checkNotNull(
+                inputPythonSourcePath.get().asFile.copyTo(
+                    enclaveDestinationPythonFile,
+                    overwrite = true
+                )
+            ) { "Python script not copied correctly while building the Gramine SGX manifest" }
+        }
+        check(signDirectManifest(manifestPath, privateKeyPath).exitValue == 0) { "Could not sign the manifest" }
+        check(sgxGetToken().exitValue == 0) { "Could not get the token for the SGX manifest" }
```
shamsasari, line 67: We should be throwing GradleException

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
bonifazi-r3: We have that just before calling such a constructor. You think it is really required?
```
        validateEnclaveMode()
        val env = GramineEnclaveEnvironment(enclaveClass, hostInterface, simulationMrSigner!!, enclaveMode)
```

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
bonifazi-r3: It is a parameter we have previously built in the manifest.
This is to make sure that we have an SGX manifest with the SGX debug flag (which is a Gramine parameter) corresponding to the enclave mode that we have in it (which is more a Conclave thing).
It might be redundant, but I like the validation bit.
I imagine a situation where we might build the sgx manifest manually and make mistake when doing it.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
shamsasari: Yes, becuase the class should do its own checking for any preconditions it has.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEntryPoint.kt
shamsasari: This is _more_ likely to introduce errors, since now we have two tightly coupled parameters.

integration-tests/python-enclave/enclave/build.gradle
bonifazi-r3: It does not run otherwise and it will ask you to increase that. We have a similar number in the Python examples in Gramine.

CC @ac101m

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
bonifazi-r3: This change allows us to run the code, otherwise it won't work.

We need a specific number of threads to run enclave code in Gramine (i.e. Python).
In addition to that we need another set of threads to handle the communication enclave/host.
With the previous code we only allowed 8 threads to be actually used inside the Enclave.
You need to consider that Gramine takes at least 3 threads for other operations, which means that we only had 5 threads to run Python stuff.

CC @ac101m

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -38,12 +45,30 @@ fs.mounts = [
...
+    "file:{{ gramine.libos }}",
+    "file:{{ gramine.runtimedir() }}/",
+    "file:{{ arch_libdir }}/",
+    "file:/usr/{{ arch_libdir }}/",
+    "file:{{ java_home }}/",
+    "file:{{ ld_preload }}",
+    "file:{{ python_packages_path }}/",
+    "file:{{ python.stdlib }}/",
+    "file:{{ python.distlib }}/",
+    "file:{{ python.get_path('stdlib', vars={'installed_base': '/usr/local'}) }}/",
```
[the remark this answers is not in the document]
bonifazi-r3: This is a copy and paste from PyTorch examples in Gramine, which I think it is a good idea.
https://github.com/gramineproject/examples/blob/master/pytorch/pytorch.manifest.template

This helped me to fix an issue in using Numpy. It should add this directory to the trusted section.
`/usr/local/lib/python3.8/dist-packages`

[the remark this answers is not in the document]
bonifazi-r3: Note that this directory seems to be specific to Ubuntu, but this is what we are using

integration-tests/python-enclave/enclave/build.gradle
shamsasari: Can you add a comment to that effect here. This means the plugin will need to have different default values depending on whether its graalvm, gramine or python. I'll create a ticket for that.

shamsasari: https://r3-cev.atlassian.net/browse/CON-1223

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
shamsasari: We shouldn't be taking the user input of `maxThreads` and then modifying it in mysterious ways hidden from the user. Instead it is better to have sensible default values that should work in most cases. See https://r3-cev.atlassian.net/browse/CON-1223. There's no need to do anything right now.

ac101m: To elaborate, because we're creating threads greedily in the call interface, some of the threads are eaten up by that (bar 8), which will potentially prevent the user from creating threads of their own.

This isn't the whole story though, I've actually discovered some additional complexity here while looking into CON-360. It turns out gramine is actually able to re-use thread control blocks when a thread yields, meaning the number of threads that can conceptually exist in the enclave isn't actually limited to the TCB count, just the number of running threads.

Not sure exactly what should be done about it, but doubling ensures that the level of concurrency available should be roughly equal to the value in the build.gradle regardless of how the user decides to do it (either creating threads in the enclave or calling the enclave concurrently from the host).

Should also probably implement your previous suggestion of not greedily creating threads/connections in the call interface!

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: Ok, I can only save a line of code for the enclave jar, is this what you mean?

integration-tests/python-enclave/signing/sample_private_key.pem
bonifazi-r3: It is used when building the SGX manifest, what would be the alternative? Don't we do similar things elsewhere?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
shamsasari: Yep.

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -38,12 +45,30 @@ fs.mounts = [
...
 ]
 
-sgx.enclave_size = "4G"
+sgx.debug = {{ sgx_debug }}
+
+{% if is_python_enclave == "true" %}
+   sgx.enclave_size = "8G"
+{% else %}
+   sgx.enclave_size = "4G"
+{% endif %}
```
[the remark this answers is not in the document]
bonifazi-r3: I think we might need an extra parameter in Conclave extension, which is somehow annoying as we have `maxHeapSize` and `maxStackSize` already. But in Gramine the `enclave_size` also includes `stack.size`. 
I will hardcode for the time being (to 4G and 8G, another different is that they need to be power of 2)

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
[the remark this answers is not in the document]
shamsasari: A comment please.

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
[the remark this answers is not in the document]
shamsasari: Doesn't that just mean we need to add the two and pass that into `enclave_size`?

integration-tests/python-enclave/signing/sample_private_key.pem
shamsasari: We do, but that might be due to other reasons, like having a stable MRSIGNER value. Here, I'm not sure we need that now. In the `release` block in the enclave build.gradle you can specify `dummyKey`.

bonifazi-r3: Ok, great. Note: I was bitten by the enclave issue we spoke about, I need to clean the enclave in order for the change to be happening and to run the test

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -194,40 +201,119 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+                    throw GradleException(
+                        "Only a single Python script is supported, but ${pythonFiles.size} were " +
+                                "found in $pythonSourcePath"
+                    )
+                }
+            }
+            task.directManifest.set(generateGramineManifestTask.manifestFile)
+            task.sgxManifest.set(Paths.get(outputSgxManifestPath).toFile())
+            task.sgxToken.set(Paths.get(outputTokenPath).toFile())
+            task.sgxSig.set(Paths.get(outputSig).toFile())
```
shamsasari, line 240: I think a single `OutputDir` would be better, but it's fine for now. I think I might end up changing this as part of the improvements to the plugin tests I'm doing on the side

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -194,40 +201,119 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+            if (pythonSourcePath != null) {
+                val pythonFiles = target.fileTree(pythonSourcePath).files
+                if (pythonFiles.size == 1) {
+                    task.pythonSourcePath.set(pythonFiles.first()!!)
+                } else {
+                    throw GradleException(
+                        "Only a single Python script is supported, but ${pythonFiles.size} were " +
+                                "found in $pythonSourcePath"
+                    )
+                }
```
shamsasari, line 235: This seems to be duplicated

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -194,40 +201,119 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
+                    SigningType.DummyKey -> createDummyKeyTask.outputKey
+                    SigningType.PrivateKey -> enclaveExtension.signingKey
+                    else -> target.provider { null }
+                }
+            })
+            task.enclaveJar.set(enclaveFatJarTask.archiveFile)
+            if (pythonSourcePath != null) {
+                val pythonFiles = target.fileTree(pythonSourcePath).files
+                if (pythonFiles.size == 1) {
+                    task.pythonSourcePath.set(pythonFiles.first()!!)
```
shamsasari, line 229: `first()` is not nullable

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
```diff
@@ -64,7 +68,8 @@ open class GenerateGramineManifest @Inject constructor(
          * Because of this, we need to add some safety margin.
          */
         val enclaveWorkerThreadCount = maxThreads.get()
-        val gramineMaxThreads = enclaveWorkerThreadCount + 8
+        //  TODO: Create a new parameters to distinguish between proper Enclave threads and threads used in the Enclave/Host Interface
```
shamsasari, line 71: That's not the right approach. We could decide to not use threads per connection, for example if we switch to using shared memory for the IPC. And so the issue is an implementation detail. I think for now link to the JIRA ticket.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
```diff
@@ -18,12 +19,14 @@ import java.security.interfaces.RSAPublicKey
...
-        private val buildType: BuildType
+open class GenerateGramineDirectManifest @Inject constructor(
+    objects: ObjectFactory,
+    private val buildType: BuildType
 ) : ConclaveTask() {
     companion object {
-        const val MANIFEST_TEMPLATE = "java.manifest.template"
+        const val MANIFEST_TEMPLATE = "$GRAMINE_MANIFEST.template"
+        const val PYTHON_ENCLAVE_SIZE = "8G"
+        const val JAVA_ENCLAVE_SIZE = "4G"
```
shamsasari, line 29: Add TODO that the enclave_size property should be the sum of the heap and stack size configs. You can probably link to this ticket: https://r3-cev.atlassian.net/browse/CON-1221

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
```diff
@@ -0,0 +1,96 @@
...
+        val enclaveDestinationPythonFileName = "${outputSGXManifestPath}/$PYTHON_FILE"
+        val enclaveDestinationPythonFile = File(enclaveDestinationPythonFileName)
+
+        if (pythonSourcePath.isPresent) {
+            checkNotNull(
+                pythonSourcePath.get().asFile.copyTo(
+                    enclaveDestinationPythonFile,
+                    overwrite = true
+                )
+            ) { "Python script not copied correctly while building the Gramine SGX manifest" }
```
shamsasari, line 65: `copyTo` returns a boolean if the copy was a success or not, rather than throwing an IOException. This odd API choice is due to the underlying `java.io.File` API, which has these issues, and this is why I always say to use the newer `Path` API whenever possible (which was created to address these issues).

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        if (signDirectManifest(manifestPath, privateKeyPath).exitValue != 0) {
```
shamsasari, line 68: Gradle by default checks the exit code for you, so there's no need to do this.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    @get:OutputFile
+    val sgxSig: RegularFileProperty = objects.fileProperty()
+
+    override fun action() {
+        val manifestPath = directManifest.get().asFile.absolutePath
+        val privateKeyPath = privateKey.get().asFile.absolutePath
+        val outputSGXManifestPath = sgxManifest.asFile.get().parentFile.absolutePath
+        val enclaveDestinationJarName = "${outputSGXManifestPath}/$GRAMINE_ENCLAVE_JAR"
+        val enclaveDestinationJarFile = File(enclaveDestinationJarName)
```
shamsasari, line 49: Does the Gramine tools require that all the input files exist in the same directory? If so then why are you passing in the absolute paths in the command line args? Shouldn't the file names suffice?

And if a common working directory is required, then use the task's [temporary directory](https://docs.gradle.org/current/dsl/org.gradle.api.Task.html#org.gradle.api.Task:temporaryDir), rather than taking one of the input files' directory.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
```diff
@@ -0,0 +1,96 @@
...
+            GRAMINE_SGX_SIGN_EXECUTABLE,
+            "--manifest=$directManifest",
+            "--key=$privateKey",
+            "--output=${sgxManifest.asFile.get().absolutePath}"
+        )
+        val enclaveJarBuildDir = sgxManifest.asFile.get().parentFile.absolutePath
+        return commandLine(command, commandLineConfig = CommandLineConfig(workingDir = enclaveJarBuildDir))
+    }
+
+    private fun sgxGetToken(): ExecResult {
```
shamsasari, line 88: What is this SGX token? There doesn't seem to be any information in their docs.

shamsasari (changes requested):

integration-tests/python-enclave/signing/sample_private_key.pem
bonifazi-r3: DOne

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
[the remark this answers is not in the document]
bonifazi-r3: I would not recommend that, `enclave_size` includes this as well, for example https://gramine.readthedocs.io/en/stable/manifest-syntax.html#gramine-internal-metadata-size

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
bonifazi-r3: Nope, I am quite against this. First of all Gramine `enclave_size` needs to be power of 2, not really sure we can easily sum two numbers and achieve that, if they are not the same power of 2 value!

shamsasari: That can be solved by just going up to the nearest power of 2.

Why is this a bad thing? We have these two configs in the enclave build.gradle and they need to be wired up in Gramine. What do you suggest we do instead?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: `copyTo` seems to return a File actually. Somehow it is so immediate to use `asFile` from the `RegularFile` property, hence I used this, which worked.
Should I use 

```

            Files.copy(enclaveJar.get().asFile.toPath(), enclaveDestinationJarPath)

``` 
 What should I use instead?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: How is this handled? I see this when I go deeper in the call stack
```
            if (result.exitValue == 0) {
                project.logger.info(execErrorStream.toString().trimEnd())

                // The non-messages have been logged and should be dumped because they are not error messages
                execErrorStream.reset()
            } else if (commandLineConfig.errorOutputStream == null) {

                // Log the error because the callee did not set an error stream
                project.logger.error(execErrorStream.toString().trimEnd())
            }

            return result
```

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
shamsasari: Yes I appreciate that Gradle uses `File` and so it's sometimes easier to use that over `Path`. I do that same as well sometimes, e.g.`writeText`. However, most times stick to `Path`, especially when you have deal with `File`'s poor API. https://www.baeldung.com/java-path-vs-file gives some more info on this.

With regards to sample, I think you can do `enclaveJar.get().asFile.toPath().copyTo(...`

Btw, why do u need to overwrite?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
shamsasari: That code uses `project.exec` and you can pass in an `ExecSpec` to configure it. By default it will throw an exception on non-zero exit codes: https://docs.gradle.org/current/javadoc/org/gradle/process/BaseExecSpec.html#isIgnoreExitValue--

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: Yes, Gramine require files to be in the same directory. I am indirectly using `gramineBuildDirectory` for that, what is wrong with that? I.e. having the files that we build in the build directory isn't somehow normal? It makes very easy to find the files that we have generated.

shamsasari: The task doesn't know that, you've added that assumption in this code. Perhaps you need an `@InputDir` instead of separate input files? Or just use the tasks temp, which exists in the project's build directory.

This isn't an major issue, and I certainly don't want to block this PR on this since I want to try this out. However I do think this isn't clean design.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: https://gramine.readthedocs.io/en/stable/devel/building.html#prepare-a-signing-key
My understanding is that this is related to the SGX EINITTOKEN. Is this something that was needed for EPID? It looks like we can get rid of it actually.
All the examples I have met are using it though.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
shamsasari: Add a TODO for the suggested solution.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
shamsasari: It does look like it's related to EPID. Remove it if that's the case.

bonifazi-r3: Removed

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
bonifazi-r3: maxHeapSize and maxStackSize are more Graal parameters than SGX related and they can be quite different.
Rounding up might waste a lot of memory.
I would suggest a Gramine specific parameter, `gramineEnclaveSize`. I feel that at some point we might need two separate sections in the `build.gradle`

ac101m: I don't like the addition of another parameter for the enclave size, maybe re-use the same parameter and warn the user that it's rounded to the next power of two, or produce an error telling them to make it a power of two?

bonifazi-r3: Which parameter would you reuse?

ac101m: Ignore my previous comment, was under the impression we had an `enclaveSize` parameter, looks like we don't!

I'd just connect it and force the user to round it up to a power of two, or print a warning that it is being done. Definitely **don't** do something like `(threadCount * stackSize) + heapSize` though.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: No need, I have removed it.
I think I vaguely remember we had cases where copy did not succeed and Gradle was silent about this.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: Readded! Getting a failure otherwise. It is not harming, so I would keep it.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
bonifazi-r3: Not sure I understand, would you use `(threadCount * stackSize) + heapSize` and round that up? Or `heapSize` and round that up?
In both cases I see risky situations and waste of almost 2^(k-1) memory.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
shamsasari: That's exactly the issue with `File` API.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
shamsasari: yes, the above `else` code from `ConclaveTask` will not execute if there's a nonz-zero value and the default for ignoreExitValue is kept to false.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
bonifazi-r3: Apologies, I don't understand.
Is there the possibility of having more stacks in the enclave than `sgx.thread_num`? My understanding is no.
What you mean by "normal Posix call"? My understanding is that in Gramine they have reimplemented them.

But if the users set the `heapSize` as power of 2 thinking that it is the actual heap and hence `enclave_size` is power of 2, wouldn't it be misleading/wrong that we don't include stack (and PAL manifest) sizes as well?
And if the users set `heapSize` to 2.1G for example , wouldn't be a waste do round up?
Rounding up `(threadCount * stackSize) + heapSize` makes more sense to me that `heapSize` only, but I honestly dislike both options, as they are confusing and misleading.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: Ok, I have just removed my checks then

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
bonifazi-r3: So, to be clear. Not all the files actually need to be in the same dir, but only `enclave.jar` and `enclave.py`, that need to be "reachable" on the same path both at build time and runtime, and I haven't found another (easier) way for them to be somewhere apart from `./` (from the point of view of the working directory during the creation of the `manifest.sgx`)

This is because when we run the hash on them they need to be, at build time, in the same path that they will be at runtime.
Theoretically, I could copy them in my `/usr/local/lib` host directory (or similar), hash them and then use the same mount in Gramine, but this would really feel odd.
So I need to copy them on the same working directory where I run `gramine-sgx-sign` so that I can just use the name `enclave.jar` and `enclave.py` (see the `java.manifest.template`)
The other files (manifest, etc) can stay wherever, and we can pass the absolute path for them.

In relation to the temporary dir: is the temporary dir actually "temporary" though?
```
There are no guarantees that the contents of this directory will be kept beyond the execution of the task
```
I really want to keep the files, as they are useful artifacts, especially to understand what is going on.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
shamsasari: I wonder if setting `remote_attestation` to DCAP will remove the need for this.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineSGXManifest.kt
shamsasari: Your reason for doing this is good, I just think we can implement this better. Not an issue for now.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
ac101m: Yes, you can have as many threads/stacks in the enclave as you like with gramine, irrespective of the value of `sgx.thread_num`. This is what I was talking about yesterday. I'd also wager that gramine does _not_ greedily allocate stack memory at startup. So no, 100% _does not make sense_ to make this a function of stack size.

As for the rounding, I think you're over-thinking this. People rarely plan the memory utilisation of their programs ahead of time. They run them, and if they run out of memory, they allocate more. So long as they are aware that the rounding is happening, or they are forced to round themselves, I'd say that's good enough.

shamsasari: >Yes, you can have as many threads/stacks in the enclave as you like with gramine, irrespective of the value of sgx.thread_num. 

Are you sure that also applies to concurrent ECALLs? Because I'm not convinced it is.

PR #85, opened 2022-11-03
CON-1137: Show version number in Conclave Init with --version option

anien:


conclave-init/src/main/kotlin/com/r3/conclave/init/cli/Main.kt
```diff
@@ -28,6 +28,7 @@ fun main(args: Array<String>) {
     headerHeading = "%n",
     synopsisHeading = "%nUsage:%n",
     optionListHeading = "%n",
+    versionProvider = ConclaveInitCli.ManifestVersionProvider::class,
```
bonifazi-r3, line 31: Question, as I don't know about this. That does `versionProvider` add practically? I.e. how is this used? Is there any convention or some automatic behaviour around?
Might be good to add description to the PR.

test-conclave-init.sh
```diff
@@ -13,6 +15,12 @@ pushd build
 
 conclaveInitJar=$(find repo/com/r3/conclave/conclave-init/ -name 'conclave-init-*jar' -not -name 'conclave-init-*javadoc.jar' -not -name 'conclave-init-*-sources.jar')
 
+conclaveInitVersion=$("$JAVA_HOME"/bin/java -jar "$conclaveInitJar" -V | awk '{print $3}')
```
bonifazi-r3, line 18: Sounds good here.
Note a thing though: while testing, I noticed that the previous line does not seem to work when there are multiple jar present. This can happen when we build the SDK multiple times and we have more versions of the jars produced.
Not related to this, might be relevant for a future fix.

bonifazi-r3 (changes requested):
Some minor comments, I also suggest to fill the description field.

conclave-init/src/main/kotlin/com/r3/conclave/init/cli/Main.kt
anien: It's required by picocli when retrieving version number dynamically (e.g. not from a const val).
I will now play with your suggestion using conclave version as a const variable as that would make it easier.

test-conclave-init.sh
anien: Okay, let me check if I can fix it quickly or we will leave it for later.

conclave-init/src/main/kotlin/com/r3/conclave/init/cli/Main.kt
anien: I have added comment to the code. versionProvider is still required.

test-conclave-init.sh
anien: Fixed. Now it only picks up the latest created/modified .jar file.

bonifazi-r3 (approved):

PR #93, opened 2022-11-09
Improve the Python README section

shamsasari:


anien (changes requested):
Looks good! 
I hope we don't forget to remove hardcoded version (1.4-SNAPSHOT) before the release. 
Just one small request not related to PR, there is missing _be_ in the sentence below:

```
This function is also much simpler than `receiveMail` in that the response can simply be returned from the function and it will **be** encrypted as a response mail back to the sender key and posted.
```

shamsasari:
PTAL

PR #95, opened 2022-11-09
CON-1187: Rename Gramine bundle and enclave-fat filenames

bon000:


shamsasari (approved):

bonifazi-r3:
@shamsasari, @anien has found out a similar issues related to another jar. Would this PR be ok for you? I have also changed the gramine jar as well to be consistent, any issues?

PR #96, opened 2022-11-11
CON-1262: Make sure EnclaveHost.close() cannot be called inside a callback

shamsasari:
We had an integration test that tested this behaviour, but this is clearly incorrect. The state of the enclave if `close()` is called in the callback after it returns back into the enclave is undefined and probably not usable.

### DIFF
--- a/conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
+++ b/conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
@@ -582,6 +582,7 @@ abstract class Enclave {
      */
     private inner class StopCallHandler : CallHandler {
         override fun handleCall(parameterBuffer: ByteBuffer): ByteBuffer? {
+            check(!lock.isHeldByCurrentThread) { "Enclave cannot be shutdown in a callback" }
             lock.withLock {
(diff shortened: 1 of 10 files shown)


shamsasari:
@ac101m The exception that the enclave gives back is ` java.lang.RuntimeException: java.lang.IllegalStateException: Enclave cannot be shutdown in a callback`, i.e the `IllegalStateException` is being wrapped in a `RuntimeException`. The only reason to wrap an exception in `RuntimeException` is to workaround checked exceptions, but that's not an issue here since `IllegalStateException` isn't checked. Any ideas where this could be happening? I've tried to look at the code and it doesn't look like it's doing this wrapping.

ac101m:
> @ac101m The exception that the enclave gives back is ` java.lang.RuntimeException: java.lang.IllegalStateException: Enclave cannot be shutdown in a callback`, i.e the `IllegalStateException` is being wrapped in a `RuntimeException`. The only reason to wrap an exception in `RuntimeException` is to workaround checked exceptions, but that's not an issue here since `IllegalStateException` isn't checked. Any ideas where this could be happening? I've tried to look at the code and it doesn't look like it's doing this wrapping.

Checked the Gramine integration tests, doesn't look like it's happening there. Might be happening in native code?

PR #98, opened 2022-11-16
CON-1178: Integrate Gramine SGX attestation

bonifazi-r3:
Here we introduce attestation for the Gramine flow.
An effort has been done in order to keep the hierarchy of classes as clean as possible with the one we currently use for Native Image/Graal.
In Gramine attestation works differently from Native Image/Graal flow.
In Native Image:

1. the host requests the "target info" (through SGX calls) and pass it to the enclave;
2. the enclave uses the "target info" to create a report, this is done with an in-enclave SGX call;
3. the enclave triggers another call to the host, asking it to generate a signed quote (here we have a recursive call);
4. the host returns the signed quote to the enclave, concluding the recursive call here above;
5. the enclave returns the signed quote back to the host
6. the host attests the signed quote calling the provisioning enclave (SGX calls)

In Gramine the flow is a bit simpler

1. the host asks for a signed quote to the enclave
2. the enclave create a report. Internally the "target info" is automatically retrieved by Gramine (by passing an empty array to the pseudo filesystem `/dev/attestation/target_info`, similarly to what we do in Native with the call `sgx_get_target_info`, where an empty array is passed in. The only difference is that in Gramine this is happening in the enclave)
3. the enclave prepares for us, automatically, a signed quote, by doing the quoting in background. There is a call to the AESM service for this.
4. the enclave returns the signed quote to the host (same as in Native Image)
5. the host attests the signed quote (same as in Native Image)

Note that the `initializeQuote` function is returning an empty `SgxTargetInfo`, which is the one passed to the enclave to create the report later on. We could also skip this step, and even remove that class and move the `Native.initQuoteDCAP` in the Attestation Service; in such a case we would need to directly pass the empty array in the `GramineEnclaveEnvironment->getSignedQuote->createReport`. I kept everything as it is just for consistency with the current code, but this can change in the futur5e.


bonifazi-r3:
I've added everyone to the review, but it is not meant to be reviewed yet, just wanted any feedback in case there is something obviously wrong or that can be improved to help @shamsasari .
Note that at the moment you need to compile a library on your VM and install/copy it in `/usr/lib` in order to make this work (and obviously you need to have the DCAP client from Azure)

```
marco@marcovm:~/conclave-core-sdk/gramine-cpp$ cmake . -Bbuild && cd build && make && sudo make install
```
Old instructions here:https://github.com/R3Conclave/conclave-core-gramine/tree/marco_gramine_attestation/hello-world-gramine#build-the-dcap-gramine-jni-layer

ac101m (commented):
I don't know much about the attestation process, but isn't the only thing that happens in the enclave the report generation? If so, can't we just get the report from Gramine and use the existing code path on the host side to do everything else, or is that too specific to native mode?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -290,7 +318,13 @@ abstract class Enclave {
...
     ): ByteCursor<SgxSignedQuote> {
         val report = env.createReport(quotingEnclaveInfo, reportData)
-        return env.hostInterface.getSignedQuote(report)
+
+        return if (env is GramineEnclaveEnvironment) {
+            val signedQuoteBytes = getQuoteFromGramine(reportData!!.bytes)
+            Cursor.wrap(SgxSignedQuote, signedQuoteBytes)
+        } else {
+            env.hostInterface.getSignedQuote(report)
+        }
```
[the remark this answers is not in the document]
ac101m: Just seems odd to be doing:

```
if (is specific subclass) {
    do specific subclass thing
}
```

bonifazi-r3:
Thanks!
There is a difference: in Gramine we get the SgxSignedQuote from the enclave. Gramine communicates with the quoting enclave to sign the report for us (in background), we don't need to do anything.

Only after that, we need the help of the host and the dcap client to retrieve the collateral and attest the quote.

So, there is no need to do a request "enclave to host" to sign the "report", hence that weird (and to improve) branch condition, where I intentionally skip the enclave to host call.


________________________________

ac101m:
> There is a difference: in Gramine we get the SgxSignedQuote from the enclave. Gramine communicates with the quoting enclave to sign the report for us (in background), we don't need to do anything.

Gotcha, I'll leave you to it ;)

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
```diff
@@ -77,6 +83,8 @@ open class GenerateGramineDirectManifest @Inject constructor(
                 "-Djava_home=${System.getProperty("java.home")}",
                 "-Darch_libdir=/lib/$architecture",
                 "-Dld_preload=$ldPreload",
+                "-Disv_prod_id=${isvProdId.get()}",
+                "-Disv_svn=${isvSvn.get() + 1}",
```
filipesoliveira, line 87: Is the plus one intended?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -196,6 +196,8 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
         return target.createTask("generateGramineManifest$type", type) { task ->
             task.pythonEnclave.set(pythonSourcePath != null)
             task.signingKey.set(signingKey)
+            task.isvProdId.set(conclaveExtension.productID)
+            task.isvSvn.set(conclaveExtension.revocationLevel)
```
filipesoliveira, line 200: Yesterday, you showed me that the revocation level is set to isvSvn - 1. Shouldn't the value be incremented by one?

gramine-cpp/jni_utils.hpp
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+    int size() const {
+        return array == nullptr ? 0 : jniEnv->GetArrayLength(array);
+    }
+};
+
+struct JniString {
+    JniString() = delete;
+    JniString(const JniString&) = delete;
```
filipesoliveira, line 38: Should the copy assignment operator be deleted too?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
bonifazi-r3: The truth is that I can't pass that value here for some reason, being this a property. I increment this in the manifest

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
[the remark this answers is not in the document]
bonifazi-r3: Changed

bonifazi-r3:
@shamsasari I have done quite a refactoring and a cleanup, struggled a bit with many details. Not sure if this deserves even a discussion as the logic of attestation in the different modes is very convoluted. Also, I am not entirely sure about a the logic of retrieving the signed quote (in the Graal flow), looks like there is a recursive call there.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/EnclaveHandle.kt
```diff
@@ -25,6 +25,15 @@ interface EnclaveHandle {
...
     val mockEnclave: Any
 
+    /**
+     * This is to establish which entity triggers the quoting: HOST or ENCLAVE.
+     * In Graal/Native Image it is responsibility of the host to use sgx_qe_get_quote and
+     * pass the result of it to the enclave.
+     * In Gramine it is responsibility of the enclave (and Gramine internals) to get the quote through AESM service.
+     * Note: Mock mode is following Graal/Native Image approach: the host does a call to the enclave to get the quote.
+     */
+    val quotingManager: QuotingManager
```
ac101m, line 35: Again, don't fully understand this.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/MockEnclaveEnvironment.kt
```diff
@@ -103,6 +103,18 @@ class MockEnclaveEnvironment(
...
     }
 
+    override fun getSignedQuote(
+        quotingEnclaveInfo: ByteCursor<SgxTargetInfo>?,
+        reportData: ByteCursor<SgxReportData>?
+    ): ByteCursor<SgxSignedQuote> {
+        //  Note that the quotingEnclaveInfo passed in is empty.
+        //  We could have directly created an empty quotingEnclaveInfo here, 
+        //    but we want to keep the same mechanism we have in the Graal/Native Image flow, as this is
+        //    what Mock mode is trying to emulate.
```
shamsasari, line 113: Not necessarily, especially now that Gramine does it differently to GraalVM. My immediate thought here is that we now have the wrong abstraction and need to tweak it.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/EnclaveHandle.kt
shamsasari: I think the one reason you've introduced this enum is bc mock mode emulates GraalVM. But I don't believe that's correct. Let it do what makes sense for it. If Gramine and GraalVM behaved the same then yes, it would make sense for mock to do the same as well. But since that's not the case, let's not force GraalVM behaviour on mock mode.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -28,103 +31,146 @@ class GramineEnclaveEnvironment(
...
+        }
     }
 
-    private val aesSealingKey by lazy(LazyThreadSafetyMode.NONE) {
-        digest("SHA-256") { update(enclaveClass.name.toByteArray()) }.copyOf(16)
+    private fun writeUserReportData(data: ByteArray) {
+
+        return FileOutputStream("/dev/attestation/user_report_data").use {
+            it.write(data)
+        }
```
shamsasari, line 125: `Paths.get("/dev/attestation/user_report_data").writeBytes(data)`

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -28,103 +31,146 @@ class GramineEnclaveEnvironment(
...
     }
 
-    override fun createReport(
-            targetInfo: ByteCursor<SgxTargetInfo>?,
-            reportData: ByteCursor<SgxReportData>?
+    private fun writeTargetInfo(data: ByteArray) {
+        FileOutputStream("/dev/attestation/target_info").use {
+            it.write(data)
+        }
```
shamsasari, line 131: similar

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceDCAP.kt
```diff
@@ -1,25 +1,23 @@
...
 import com.r3.conclave.host.internal.NativeLoader
 import java.nio.ByteBuffer
 
-class EnclaveQuoteServiceDCAP(val attestationParameters: AttestationParameters.DCAP): EnclaveQuoteService() {
+class EnclaveQuoteServiceDCAP : EnclaveQuoteService() {
 
     override fun initializeQuote(): Cursor<SgxTargetInfo, ByteBuffer> {
         val targetInfo = Cursor.allocate(SgxTargetInfo)
-        Native.initQuoteDCAP(NativeLoader.libsPath.toString(), targetInfo.buffer.array())
+        Native.initQuoteDCAP(NativeLoader.libsPath.toString(), false, targetInfo.buffer.array())
```
shamsasari, line 12: I think the parameter should be "load", not "skip". I initially though this is skipping the DCAP lib.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceGramineDCAP.kt
```diff
@@ -0,0 +1,22 @@
...
+        //  When using Gramine, we do not need to load (dlsym) all the quoting libraries,
+        //    hence we pass a true bool value to avoid loading them.
+        Native.initQuoteDCAP(NativeLoader.libsPath.toString(), true, targetInfo.buffer.array())
+        return targetInfo
+    }
+
+    override fun retrieveQuote(report: ByteCursor<SgxReport>): ByteCursor<SgxSignedQuote> {
+        //  This is not executed in the context of Gramine flow.
+        throw IllegalStateException("The quote can't be retrieved from the host when running Gramine")
+    }
```
shamsasari, line 21: This tells us where the abstraction no longer is fit for purpose. This `retrieveQuote` needs to be an implementation detail of GraalVM. We need an abstraction which is "higher level" to nicely fit Gramine into.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
shamsasari: `revocationLevel` is isvSvn + 1, either do that here, or rename `task.isvSvn` to `revocationLevel` and do it in the . I think the later is the correct way.

shamsasari (changes requested):

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -271,26 +271,13 @@ abstract class Enclave {
      * @Throws IllegalArgumentException If the provided report data byte array is present, but is not 64 bytes long.
      */
     @Beta
-    fun createAttestationQuote(reportData: ByteArray?): ByteArray {
-        val reportDataCursor = reportData?.let {
-            require(reportData.size == 64) {
-                "User report data must be 64 bytes long, but was ${reportData.size} bytes instead."
-            }
-            ByteCursor.wrap(SgxReportData, reportData)
+    fun createAttestationQuote(reportData: ByteArray): ByteArray {
```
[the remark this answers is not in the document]
bonifazi-r3: I just noticed that this is used by our integration tests. Is this something we expose to users and it is used?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -28,103 +31,146 @@ class GramineEnclaveEnvironment(
...
+    ): ByteCursor<SgxReport> {
+        return if (enclaveMode == EnclaveMode.SIMULATION) {
+            createSimulationReport(reportData)
+        } else {
+            val report = retrieveReport(
+                targetInfo?.buffer?.getRemainingBytes(avoidCopying = true) ?: byteArrayOf(),
+                reportData?.buffer?.getRemainingBytes(avoidCopying = true) ?: byteArrayOf()
+            )
+            Cursor.slice(SgxReport, ByteBuffer.wrap(report))
+        }
```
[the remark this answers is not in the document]
bonifazi-r3: Ok, great. This is somehow what we noticed since introducing Gramine, simulation mode is quite different from debug and release. Will split in two classes

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
```diff
@@ -28,103 +31,146 @@ class GramineEnclaveEnvironment(
...
+    }
 
-    private val currentCpuSvn: ByteArray by lazy {
-        versionToCpuSvn(tcbLevel)
+    override fun getSignedQuote(
+        quotingEnclaveInfo: ByteCursor<SgxTargetInfo>?,
+        reportData: ByteCursor<SgxReportData>?
```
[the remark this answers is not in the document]
bonifazi-r3: Will revert in the beta API

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineEnclaveEnvironment.kt
bonifazi-r3: Always forget, doing it

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/MockEnclaveEnvironment.kt
bonifazi-r3: I have noticed that in Mock mode we pass an empty `quotingEnclaveInfo`.
What do you mean by "not necessarily"? I haven't changed mock mode, just refactored functions to follow the common interface.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/EnclaveHandle.kt
bonifazi-r3: I have introduced this to avoid the ugly `is NativeHandle` vs `is GramineHandle`, to distinguish between the flow to get the "signed quote" when working in Graal vs Gramine.

Mock mode is somehow untouched by my changes, it is doing what was doing before, i.e. following the same approach as Graal.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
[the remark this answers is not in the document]
shamsasari: Of course it is. It's in a public member in a public package (i.e. non-internal).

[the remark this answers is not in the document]
shamsasari: It's used by the KDS when it gets the hsm master key from AKV

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceDCAP.kt
bonifazi-r3: It is skipping the quoting part of it actually. Would `loadQuotingLibraries` and revert the logic be fine?

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/MockEnclaveEnvironment.kt
shamsasari: >What do you mean by "not necessarily"?

That mock mode should emulate GraalVM, at least now that we have Gramine doing differently.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceDCAP.kt
shamsasari: Yes, I think it's clearer

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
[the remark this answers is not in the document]
ac101m: For some context, the use of this is to create signed quotes that include some enclave-internal runtime data (like the hash of a public key). This is how clients create sealed blobs that only the enclave can decrypt (e.g. for mail). And is also how AKV secure release works. So yeah, absolutely needs to be here!

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/MockEnclaveEnvironment.kt
bonifazi-r3: Sorry, I don't understand the point.
I have just added a comment here, there are no functional changes of the mock mode flow, just the implementation.
Mock mode WAS emulating Graal VM and it IS emulating Graal VM way of signing the quote (or better that ping-pong host-enclave mechanism). What do I need to fix?
Or did I change anything without noticing?

shamsasari: I understand you've not change anything with regards to mock mode behaviour. I guess what I'm saying is, now that Gramine has been introduced and it does things differently, shouldn't mock mode also not be changed to do this differently, or rather more simply? There's no reason now why mock mode should emulate graalvm. But I guess that is out of scope of this PR and can be done later? The reason I think it could be done here is if we also redesigned the host's attestation abstraction in this PR then we can simply mock mode attestaion as well.

I guess this can be all done in a separate PR. I guess add a TODO about changing the abstaction and simplifying mock mode attestsation.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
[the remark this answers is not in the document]
bonifazi-r3: Ok, sorry, I think I misunderstood the @Beta thing, I thought that this was some kind of deprecated code and I also sloppily followed the IntelliJ suggestion that this was always non-null (as we have only 1 integration tests with non-null argument calling this).
I will revert, thanks.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/MockEnclaveEnvironment.kt
bonifazi-r3: Ah ok yes, but then what should Mock mode emulate? Graal or Gramine or both? I just thought that emulating the most complex flow was more relevant. If not, it would imply that we could have simplified things before, then, correct?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceGramineDCAP.kt
ac101m: There is some awkwardness here because of the how the call interface abstraction works. We have a fixed enum for all modes, but different calls are available (for example, the signed quote call is available in gramine despite not being needed). So as it stands we're pretty much bound to have some odds and ends like this.

I have some thoughts about how this can be dealt with, but it's a non trivial refactor.

bonifazi-r3: The `nicely` is the culprit here :-) Yes, with Gramine we break assumptions, specifically the fact that we have a fixed number of "handlers" to be registered and used.

We can surely refactor things around but I always wondering if we need continuous refactoring for every PR/feature we want to add, as this has always have impact to the speed of delivering features useful for users vs cleanness of the code, which is somehow not immediately needed (considering that, if things go well, we could even demise Graal at some point, and another round of refactoring will be needed).

In this case I think we can even get rid of this class for Gramine, as the initialization of the native libraries can be moved into the Gramine related attestation class.
But then we would need to have some conditional code around, in order to call `initializeQuote` in non-Gramine cases, which also mean that we would need to change the way we register the handlers, as in Gramine we now skip a couple of handlers (1 host 1 enclave).
We would also need to do a not small refactoring around, I think, for example `enclaveInterface` is mapped one to one with `enclaveHandle`, so they might need to be merged/improved.

We can fit everything in this PR, but then it would be more a big refactoring where Gramine attestation would become just a minor thing. I think we should be more granular in doing this massive changes, I'd suggest to find a compromise and postpone the refactoring in a separate PR.

ac101m: I think this is a pretty simple problem to fix at this stage as this is all still fresh in our memories. Not sure if it should be done in this PR though. Definitely best to keep this restricted to attestation.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/MockEnclaveEnvironment.kt
shamsasari: Since Graal and Gramine do things slightly different from each other, I think the solution for mock is always to do things in its own way that makes most sense. This should probably be done as part of fixing the attestation abstraction, which can occur in a followup ticket.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
```diff
@@ -28,103 +30,86 @@ class GramineEnclaveEnvironment(
...
-        digest("SHA-256") { update(enclaveClass.name.toByteArray()) }.copyOf(16)
-    }
-
     override fun createReport(
-            targetInfo: ByteCursor<SgxTargetInfo>?,
-            reportData: ByteCursor<SgxReportData>?
+        targetInfo: ByteCursor<SgxTargetInfo>?,
+        reportData: ByteCursor<SgxReportData>?
     ): ByteCursor<SgxReport> {
+        check(simulationMrsigner != null) { "Simulation MRSigner not provided in the enclave" }
```
shamsasari, line 37: There's no reason now for simulationMrsigner to be nullable.

Likewise, this class will always represent simulation mode, there's no need to pass in the mode into the c'tor.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceFactory.kt
```diff
@@ -1,13 +1,19 @@
 package com.r3.conclave.host.internal.attestation
 
 import com.r3.conclave.host.AttestationParameters
+import com.r3.conclave.host.internal.QuotingManager
 
 object EnclaveQuoteServiceFactory {
-    fun getService(attestationParameters: AttestationParameters?): EnclaveQuoteService {
+    fun getService(attestationParameters: AttestationParameters?, quotingManager: QuotingManager): EnclaveQuoteService {
```
shamsasari, line 7: I would actually get rid of this class and move this method into `EnclaveHost` and then use the instance of the private `enclaveHandle` to determine if gramine or non-gramine DCAP should be used. This will remove the need to have the QuotingManager enum.

bonifazi-r3: Interesting. I actually added the enum to AVOID the `is GramineEnclaveHandle`, which seems a bit hacky to me. I will add this condition then

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
bonifazi-r3: Correct, this is a leftover of the copy of the class, removing it

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineSGXEnclaveEnvironment.kt
```diff
@@ -0,0 +1,115 @@
...
+    companion object {
+        private fun versionToCpuSvn(num: Int): ByteArray {
+            return digest("SHA-256") {
+                update(ByteBuffer.allocate(2).putShort(num.toShort()).array())
+            }.copyOf(SgxCpuSvn.size)
+        }
+    }
+
+    init {
+        require(enclaveMode != EnclaveMode.MOCK) { "Gramine can't run in MOCK mode" }
```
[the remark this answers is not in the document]
bonifazi-r3: Done

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineDirectEnclaveEnvironment.kt
bonifazi-r3: Removed that check

bonifazi-r3: But I added one in the class where this is called, to verify that the signer has been retrieved correctly, let me know if this is fine

shamsasari (changes requested):

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceGramineDCAP.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import com.r3.conclave.common.internal.*
+import com.r3.conclave.host.internal.Native
+import com.r3.conclave.host.internal.NativeLoader
+import java.nio.ByteBuffer
+
+class EnclaveQuoteServiceGramineDCAP : EnclaveQuoteService() {
+
+    override fun getQuotingEnclaveInfo(): Cursor<SgxTargetInfo, ByteBuffer> {
+        val targetInfo = Cursor.allocate(SgxTargetInfo)
```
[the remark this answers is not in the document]
shamsasari: I think the answer is no, because we don't need the target info for Gramine. Correct? If so then we can move the `initQuoteDCAP` into the companion object so that it's only loaded in once.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceDCAP.kt
```diff
@@ -1,25 +1,26 @@
...
 
-class EnclaveQuoteServiceDCAP(val attestationParameters: AttestationParameters.DCAP): EnclaveQuoteService() {
+class EnclaveQuoteServiceDCAP : EnclaveQuoteService() {
+    private val targetInfo = Cursor.allocate(SgxTargetInfo)
 
-    override fun initializeQuote(): Cursor<SgxTargetInfo, ByteBuffer> {
-        val targetInfo = Cursor.allocate(SgxTargetInfo)
-        Native.initQuoteDCAP(NativeLoader.libsPath.toString(), targetInfo.buffer.array())
+    init {
+        Native.initQuoteDCAP(NativeLoader.libsPath.toString(), false, targetInfo.buffer.array())
```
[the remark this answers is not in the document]
bonifazi-r3: The previous code was actually doing it each time we were updating the attestation and luckily we did not have issues :-)
This is meant to be called only once, would you prefer to move it to a companion object anyway?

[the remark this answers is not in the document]
shamsasari: I presume we haven't had issues because we've not tested created two `EnclaveHost` instances in the same process. So yes we should do it, but there's no need to do it now, as you this is how it was before. I also noticed that this class can be an `object` and so we get this singleton loading for free if we just change that.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceGramineDCAP.kt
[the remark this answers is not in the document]
bonifazi-r3: Yes, I am still trying to make sense of this mechanism. This class is actually useless, because in the case of Gramine I could initialize the library in the related attestation class.
The actual value of targetInfo should be null, as we should not pass any target in the context of calling `getSignedQuote` in Gramine (this is not done right now). This is because Gramine retrieves the enclave info of the Quoting enclave automatically.

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -14,8 +14,11 @@ loader.log_level = "error"
 # TODO: remove this when in production, see: https://gramine.readthedocs.io/en/stable/manifest-syntax.html#command-line-arguments
 loader.insecure__use_cmdline_argv = true
 
-# TODO: Enable DCAP attestation
-sgx.remote_attestation = "none"
+sgx.remote_attestation = "dcap"
+sgx.ra_client_spid = ""
```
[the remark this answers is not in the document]
bonifazi-r3: This is an epid parameter (which is filled with meaningful values when in EPID) which often pops up in some example together with `sgx.remote_attestation = "dcap"`
I actually tested this in order to check whether there was any difference in using this empty string parameter vs removing it. I could not spot any difference, so we can remove it.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/GramineSGXEnclaveEnvironment.kt
```diff
@@ -0,0 +1,119 @@
...
+        persistentMountPath: String,
+        encryptionKey: ByteArray
+    ) {
+        //  TODO: Gramine filesystem support
+    }
+
+    private fun retrieveReport(targetInfoBytes: ByteArray?, userReportDataBytes: ByteArray?): ByteArray {
+        if (targetInfoBytes != null) {
+            writeTargetInfo(targetInfoBytes)
+        }
```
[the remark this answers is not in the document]
shamsasari: >user report data must be written into /dev/attestation/user_report_data and target info must be written into /dev/attestation/target_info. Otherwise the obtained report will contain incorrect or stale user report data and target info.

This is what I'm talking about. Even if the user doens't provide values for these, the resulting report must not use old values. It's insecure.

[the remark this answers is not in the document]
shamsasari: This code will also suffer from race conditions if multiple threads try to get a report. Annotate this method with `@Synchronized`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/internal/MockEnclaveEnvironment.kt
bonifazi-r3: Is this ok now? If so, please mark as resolved

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -481,12 +469,15 @@ abstract class Enclave {
...
+        private var _quotingEnclaveInfo: ByteCursor<SgxTargetInfo>? = null
         private var _mostRecentQuote: ByteCursor<SgxSignedQuote>? = null
+        val quotingEnclaveInfo: ByteCursor<SgxTargetInfo> get() = checkNotNull(_quotingEnclaveInfo)
         val mostRecentQuote: ByteCursor<SgxSignedQuote> get() = checkNotNull(_mostRecentQuote)
 
         override fun handleCall(parameterBuffer: ByteBuffer): ByteBuffer? {
-            val quotingEnclaveInfo = ByteCursor.slice(SgxTargetInfo, parameterBuffer)
-            val quote = createAttestationQuote(quotingEnclaveInfo, createEnclaveInstanceInfoReportData())
+            _quotingEnclaveInfo = ByteCursor.slice(SgxTargetInfo, parameterBuffer)
+            print("GetEnclaveInstanceInfoQuoteCallHandler _quotingEnclaveInfo $_quotingEnclaveInfo")
```
shamsasari, line 479: delete

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/attestation/EnclaveQuoteServiceDCAP.kt
```diff
@@ -1,25 +1,33 @@
...
+            true,
+            it.buffer.array()
+        )
+    }
 
-    override fun initializeQuote(): Cursor<SgxTargetInfo, ByteBuffer> {
-        val targetInfo = Cursor.allocate(SgxTargetInfo)
-        Native.initQuoteDCAP(NativeLoader.libsPath.toString(), targetInfo.buffer.array())
+    override fun getQuotingEnclaveInfo(): Cursor<SgxTargetInfo, ByteBuffer> {
+        println("EnclaveQuoteServiceDCAP info $targetInfo")
```
shamsasari, line 22: delete

integration-tests/python-enclave/host/src/test/kotlin/com/r3/conclave/integrationtests/python/enclave/PythonEnclaveTest.kt
```diff
@@ -17,7 +17,7 @@ class PythonEnclaveTest {
         fun check() {
             gramineOnlyTest()
         }
-    }
+    }c
```
shamsasari, line 20: delete

shamsasari (approved):
A couple of print statements to delete, but otherwise this looks good.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
bonifazi-r3: SOrry, this was to debug an integration test. Deleted now

integration-tests/python-enclave/host/src/test/kotlin/com/r3/conclave/integrationtests/python/enclave/PythonEnclaveTest.kt
bonifazi-r3: SOrry, this was to debug an integration test. Deleted now

bonifazi-r3 (commented):
.

PR #106, opened 2022-11-25
[CON-360] exception is thrown if too many threads are requested

ac101m:
Changes:
- Re-enable disabled integration test.
- Remove unneeded error message from pthread patch.
- Bump graalvm version to 1.4-SNAPSHOT

### DIFF
--- a/integration-tests/general/tests/src/test/kotlin/com/r3/conclave/integrationtests/general/tests/ThreadSafeEnclaveTests.kt
+++ b/integration-tests/general/tests/src/test/kotlin/com/r3/conclave/integrationtests/general/tests/ThreadSafeEnclaveTests.kt
@@ -94,8 +94,6 @@ class ThreadSafeEnclaveTests : AbstractEnclaveActionTest("com.r3.conclave.integr
         assertThat(result).isEqualTo((n * (n + 1)) / 2)
     }
 
-    @Disabled(": CON-360")
-    // fatal error 'PosixJavaThreads.start0: pthread_create'
     @Test
     fun `exception is thrown if too many threads are requested`() {
         val n = 15 // > defaultTCSNum(10)


shamsasari:
@bonifazi-r3 or @filipesoliveira better placed to review the C++ code.

filipesoliveira:
Is the description of this PR correct?

Which integration test was re-enabled? And which error message was removed?

The code looks fine to me.

filipesoliveira (approved):
Code looks fine.

integration-tests/general/tests/src/test/kotlin/com/r3/conclave/integrationtests/general/tests/ThreadSafeEnclaveTests.kt
```diff
@@ -89,16 +89,17 @@ class ThreadSafeEnclaveTests : AbstractEnclaveActionTest("com.r3.conclave.integr
...
     @Test
     fun `exception is thrown if too many threads are requested`() {
-        val n = 15 // > defaultTCSNum(10)
+        // This test hangs when tearing down the enclave, so we disable teardown here.
+        // TODO: CON-1244 Figure out why this test hangs, then remove this teardown logic.
+        doEnclaveTeardown = false
+        val n = 15 // > thread safe enclave maxThreads (10)
         assertThatThrownBy {
             callEnclave(TooManyThreadsRequestedAction(n))
         }.hasMessageContaining("The enclave ran out of TCS slots when calling from a new thread into the enclave.") // SGX_ERROR_OUT_OF_TCS
```
bonifazi-r3, line 105: Where in the code we now print this message `The enclave ran out of TCS slots when calling from a new thread into the enclave.`?

cpp/linux-sgx/pthread_changes.diff
```diff
@@ -360,9 +360,6 @@ index df3feeb..f2d2fdb 100644
 +            break;
 +        }
 +        if (!m_thread_blocking || !m_thread_mutex.wait(m_deadlock_timeout)) {
-+            std::cout << "The enclave ran out of TCS slots when calling from a new thread into the enclave. Please either increase "
```
bonifazi-r3, line 363: If we remove these `cout`, there might be a `iostream` header to be removed from the includes?

integration-tests/general/tests/src/test/kotlin/com/r3/conclave/integrationtests/general/tests/ThreadSafeEnclaveTests.kt
ac101m: https://github.com/R3Conclave/graal-patches

PR #107, opened 2022-11-28
CON-1223 - Updated the default values of maxThreads and the enclave_worker_threads.

filipesoliveira:
The default value of `enclave_worker_threads` is set to 10.

Gramine's maximum threads will no longer be twice the value set in the `maxThreads` but instead the exact value provided in the `maxThreads` configuration field.

`maxThreads` default value has been set to 100. This value should cover most of the use-cases and it does not degrade the enclave performance nor does increases the memory usage of an enclave.

Removed the `maxThreads` configuration field from `integration-tests/python-enclave/enclave/build.gradle`

The docs were updated to include the new default value set for the maxThreads.

### DIFF
--- a/docs/docs/api-changes.md
+++ b/docs/docs/api-changes.md
@@ -7,6 +7,10 @@
 For security reasons, a client that sets the security level in the [`EnclaveConstraint`](api/-conclave%20-core/com.r3.conclave.common/-enclave-constraint/index.html) to `INSECURE` will no longer be able to connect to `STALE` or `SECURE` enclaves.
 The behavior of the security levels `STALE`, and `SECURE` has not changed.
 
+The default value for the Conclave configuration field `maxThreads` has been increased to 100. This value should be suitable
+to cover most of the use-cases when the logic inside the enclaves uses a large number of threads. There should be no
+considerable performance impact.
+
 ## 1.2 to 1.3
 
 ### Maven Central
--- a/docs/docs/api-changes.md
+++ b/docs/docs/api-changes.md
@@ -9,7 +9,7 @@ The behavior of the security levels `STALE`, and `SECURE` has not changed.
 
 The default value for the Conclave configuration field `maxThreads` has been increased to 100. This value should be suitable
 to cover most of the use-cases when the logic inside the enclaves uses a large number of threads. There should be no
-considerable performance impact.
+major performance impact.
 
 ## 1.2 to 1.3
 
--- a/docs/docs/enclave-configuration.md
+++ b/docs/docs/enclave-configuration.md
@@ -45,7 +45,7 @@ conclave {
     enablePersistentMap = false
     maxPersistentMapSize = "16m"
     maxThreads = 100
-    supportLanguages = ""   
+    supportLanguages = ""
     reflectionConfigurationFiles.from("config.json")
     serializationConfigurationFiles.from("serialization.json")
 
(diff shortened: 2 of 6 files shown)


shamsasari:
I'm not really in favour of exposing `enclave_worker_threads` to the user. It's really a short-term implementation detail of the Gramine process communication channel. See the TODO in `SocketEnclaveHostInterface`

ac101m:
> I'm not really in favour of exposing `enclave_worker_threads` to the user. It's really a short-term implementation detail of the Gramine process communication channel. See the TODO in `SocketEnclaveHostInterface`

Without a separate option, how will users be aware that some of their threads are being used internally by conclave?

shamsasari:
I don't see that as an issue, since we can explain that there is an overhead of threads which is used by Conclave, JVM, Gramine, etc. And since now we're setting this value to be quite high, the overhead should be proportionally small.

ac101m:
> I don't see that as an issue, since we can explain that there is an overhead of threads which is used by Conclave, JVM, Gramine, etc. And since now we're setting this value to be quite high, the overhead should be proportionally small.

So what should the call interface concurrency be set to? A constant value?

shamsasari:
I think so, for now at least. Going forward, we want to address that TODO, and create a new socket connection for each new thread, using ThreadLocals.

docs/docs/enclave-configuration.md
```diff
@@ -44,8 +44,8 @@ conclave {
     persistentFileSystemSize = "0m"
     enablePersistentMap = false
     maxPersistentMapSize = "16m"
-    maxThreads = 10
-    supportLanguages = ""
+    maxThreads = 100
+    supportLanguages = ""
```
ac101m, line 48: Got a bit of extraneous whitespace at the end of this line!

ac101m (commented):
A trivial problem

shamsasari (changes requested):
Looking good. Two things:

1. Add an entry to api-changes.md, with a small explanation for why, and say that it doesn't affect enclave performance much.
2. The PR title is not reflective of the actual changes, it's just mirroring the original JIRA ticket.

docs/docs/api-changes.md
```diff
@@ -7,6 +7,10 @@
 For security reasons, a client that sets the security level in the [`EnclaveConstraint`](api/-conclave%20-core/com.r3.conclave.common/-enclave-constraint/index.html) to `INSECURE` will no longer be able to connect to `STALE` or `SECURE` enclaves.
 The behavior of the security levels `STALE`, and `SECURE` has not changed.
 
+The default value for the Conclave configuration field `maxThreads` has been increased to 100. This value should be suitable
+to cover most of the use-cases when the logic inside the enclaves uses a large number of threads. There should be no
+considerable performance impact.
```
shamsasari, line 12: considerable -> major

shamsasari (approved):
Thanks

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineDirectManifest.kt
```diff
@@ -87,8 +79,8 @@ open class GenerateGramineDirectManifest @Inject constructor(
             "-Dpython_packages_path=$pythonPackagesPath",
             "-Dis_python_enclave=${pythonEnclave.get()}",
             "-Denclave_mode=${buildType.name.uppercase()}",
-            "-Denclave_worker_threads=$enclaveWorkerThreadCount",
-            "-Dgramine_max_threads=$gramineMaxThreads",
+            "-Denclave_worker_threads=10",
```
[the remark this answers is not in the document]
filipesoliveira: The default value was set to 10 so I decided to keep it. I don't know why we picked that value.

PR #108, opened 2022-11-28
[CON-1135] Gradle plugin to output enclave metadata to file

ac101m:
# Changes
- Output mrsigner and mrenclave files during the build

### DIFF
--- a/plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
+++ b/plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
@@ -1,8 +1,10 @@
 package com.r3.conclave.plugin.enclave.gradle
 
+import com.r3.conclave.utilities.internal.toHexString
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.OutputFile
 import org.gradle.internal.os.OperatingSystem
 import javax.inject.Inject
 import kotlin.io.path.absolutePathString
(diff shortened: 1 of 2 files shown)


plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
```diff
@@ -40,6 +53,10 @@ open class GenerateEnclaveMetadata @Inject constructor(
         }
 
         val enclaveMetadata = EnclaveMetadata.parseMetadataFile(metadataFile)
+
+        mrsignerOutputFile.asFile.get().writeText(enclaveMetadata.mrsigner.bytes.toHexString().lowercase())
```
bonifazi-r3, line 57: Any idea of what is the different between `.get().asFile` vs `.asFile.get()`? I see we use both.

bonifazi-r3: Also, why lowercase? I think we generally used uppercase for these?

bonifazi-r3 (changes requested):

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
```diff
@@ -13,9 +15,20 @@ open class GenerateEnclaveMetadata @Inject constructor(
...
+        const val ENCLAVE_MRENCLAVE_FILE = "mrenclave.txt"
+    }
+
     @get:InputFile
     val inputSignedEnclave: RegularFileProperty = objects.fileProperty()
 
+    @OutputFile
+    val mrsignerOutputFile: RegularFileProperty = objects.fileProperty()
+
+    @OutputFile
```
shamsasari, line 29: These need to have the `get:` prefix, otherwise don't work. See the Javadocs for more info.

shamsasari: Also, any reason why two files and not one. We can add further metdata to the single file in future

ac101m: So the files can be quickly added to command lines like so `$(cat <filename>)`, we don't want any parsing to be necessary.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
ac101m: Secure release process uses lowercase

ac101m:
> Correct, this needs to be raised on release/1.3

Will re-raise

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
shamsasari: Secure release policy is an Azure thing, and since we already print these values in the attesation report toString and else in uppercase, we should be consistent and do the same here.

PR #109, opened 2022-12-01
[CON-1135] Gradle plugin to output enclave metadata to file

ac101m:
# changes
- Add mrsigner and mrenclave output files
- Fix some indentation in the gradle plugin

### DIFF
--- a/docs/docs/release-notes.md
+++ b/docs/docs/release-notes.md
@@ -1,5 +1,9 @@
 # Release notes
 
+## 1.3.1
+1. To make deploying enclaves built with conclave easier, files containing the enclave code hash and signer are now 
+   generated when an enclave is built. See [here]() for more information.
+
 ## 1.3
 
 1. :tada: **The Conclave Core SDK is now open source!** :tada: Read our
--- a/docs/docs/release-notes.md
+++ b/docs/docs/release-notes.md
@@ -2,7 +2,7 @@
 
 ## 1.3.1
 1. To make deploying enclaves built with conclave easier, files containing the enclave code hash and signer are now 
-   generated when an enclave is built. See [here]() for more information.
+   generated when an enclave is built. See [here](enclave-configuration.md#enclave-build-process) for more information.
 
 ## 1.3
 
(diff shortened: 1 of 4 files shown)


ac101m:
@shamsasari @bonifazi-r3 PTAL.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
```diff
@@ -13,9 +15,20 @@ open class GenerateEnclaveMetadata @Inject constructor(
     private val buildType: BuildType,
     private val linuxExec: LinuxExec
 ) : ConclaveTask() {
+    companion object {
+        const val ENCLAVE_MRSIGNER_FILE = "mrsigner.txt"
+        const val ENCLAVE_MRENCLAVE_FILE = "mrenclave.txt"
+    }
```
shamsasari, line 21: Minor: I don't see the need for these constants, since the location of the files are defined by the plugin class.

shamsasari (changes requested):
The release notes need updating, and what about docs? Is there somewhere we can mention this.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GenerateEnclaveMetadata.kt
ac101m: Removed

shamsasari:
We dogfood the SDK with the KDS, so we shouldn't ve in the mindset that we are adding KDS-only features to the SDK, but rather it is indicative of what external users will face.

docs/docs/release-notes.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## 1.3.1
+1. To make deploying enclaves built with conclave easier, files containing the enclave code hash and signer are now 
+   generated when an enclave is built. See [here]() for more information.
```
shamsasari, line 5: You forgot to add in the `enclave-configuration.md` link

shamsasari:
@JohnJoser3 docs changes OK?

JohnJoser3:
> @JohnJoser3 docs changes OK?

Looks good to me.

JohnJoser3 (commented):
Looks good to me.

shamsasari (approved):
Looks good.

PR #110, opened 2022-12-02
CON-1164: Remove EPID and out-of-kernel driver from docs

JohnJoser3:
- Edited/deleted a few docs to remove the mentions about EPID and out-of-kernel support.
- Added instructions to use Linux 5.11 kernel or higher.
- Pointed users who want to try on-premise deployment to contact us on discord.


docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
 As cloud providers run caches, DCAP supports vendor-specific plugins. Intel provides a default one
 which requires a [subscription](https://api.portal.trustedservices.intel.com/provisioning-certification).
 
-Azure provides a DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) that does not require a subscription. Conclave
-bundles and uses Azure's DCAP plugin by default. The Azure caches are open to the public internet and can be used from anywhere. Azure Confidential Computing instances like DC4s_v3 come pre-provisioned for DCAP and as Conclave
+Azure provides a DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) that does not require a subscription. 
+Conclave bundles and uses Azure's DCAP plugin by default. The Azure caches are open to the public internet and can
```
shamsasari, line 13: remove "bundles and"

docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
...
 which requires a [subscription](https://api.portal.trustedservices.intel.com/provisioning-certification).
 
-Azure provides a DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) that does not require a subscription. Conclave
-bundles and uses Azure's DCAP plugin by default. The Azure caches are open to the public internet and can be used from anywhere. Azure Confidential Computing instances like DC4s_v3 come pre-provisioned for DCAP and as Conclave
+Azure provides a DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) that does not require a subscription. 
+Conclave bundles and uses Azure's DCAP plugin by default. The Azure caches are open to the public internet and can 
+be used from anywhere. Azure Confidential Computing instances like DC4s_v3 come pre-provisioned for DCAP and as Conclave
 comes with the necessary libraries bundled, you don't need to do any further setup.
 
+DCAP is also known as ECDSA attestation.
```
shamsasari, line 17: This should be right at the beginning in the first paragraph

docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
...
+
 ### DCAP Plugin
-To perform attestation using DCAP, Conclave needs a way to gather information about the platform the enclave is hosted on. This information provides proof from Intel that the system supports SGX and that it is patched and up to date.
+To perform attestation using DCAP, Conclave needs a way to gather information about the platform the enclave is
```
shamsasari, line 20: This paragraph seems to be a better introduction to DCAP than the existing one above. Either case, it doesn't belong here as it's duplicating what's been said above.

docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
...
+DCAP is also known as ECDSA attestation.
+
 ### DCAP Plugin
-To perform attestation using DCAP, Conclave needs a way to gather information about the platform the enclave is hosted on. This information provides proof from Intel that the system supports SGX and that it is patched and up to date.
+To perform attestation using DCAP, Conclave needs a way to gather information about the platform the enclave is 
+hosted on. This information provides proof from Intel that the system supports SGX and that it is patched and up to 
+date.
+
+DCAP is designed to work on different server topologies. So, rather than directly connecting to Intel services to 
+retrieve this information, the cloud vendor or the owner of the SGX system must provide a DCAP client plugin with
```
shamsasari, line 25: > So, rather than directly connecting to Intel services to 
retrieve this information

This sentence is implying the EPID protocol, and so I think we should bring that to attention here and use it to mention that we don't support it. So something like

"So, rather than directly connecting to Intel services to retrieve this information, like in the old EPID protocol which is no longer supported, the cloud vendor ..."

docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
...
+
 ### DCAP Plugin
-To perform attestation using DCAP, Conclave needs a way to gather information about the platform the enclave is hosted on. This information provides proof from Intel that the system supports SGX and that it is patched and up to date.
+To perform attestation using DCAP, Conclave needs a way to gather information about the platform the enclave is 
+hosted on. This information provides proof from Intel that the system supports SGX and that it is patched and up to 
+date.
+
+DCAP is designed to work on different server topologies. So, rather than directly connecting to Intel services to 
+retrieve this information, the cloud vendor or the owner of the SGX system must provide a DCAP client plugin with 
+the required information. Intel provides a generic DCAP client plugin as part of the DCAP runtime. To use this, you
```
shamsasari, line 26: To use this -> To use their plugin

docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
...
+then additional work may be required to provide data in the correct format for Conclave. Please contact R3 support 
+if you need to set this up.
 
-DCAP is designed to work on many different server topologies. So, rather than directly connecting to Intel services to retrieve this information, the cloud vendor or the owner of the SGX system must provide a DCAP client plugin with the required information. Intel provides a generic DCAP client plugin as part of the DCAP runtime. To use this, you also need to set up a Provisioning Certificate Caching Service (PCCS). Intel provides an example and some instructions [here](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/pccs/README.md). If you would like to use Intel's reference implementation of their [PCCS service](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/pccs), then additional work may be required to provide data in the correct format for Conclave. Please contact R3 support if you need to set this up.
+Microsoft has written a DCAP client plugin that works with its Azure Confidential Compute virtual machines. In fact,
```
shamsasari, line 33: Microsoft has also

docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
...
+Microsoft has written a DCAP client plugin that works with its Azure Confidential Compute virtual machines. In fact, 
+it also works outside of Azure for single CPU systems, but this may not always be the case.
 
-Microsoft has written a DCAP client plugin that works with its Azure Confidential Compute virtual machines. In fact, it also works outside of Azure for single CPU systems, but this may not always be the case.
+Conclave's bundled Azure client plugin will *only* be used if no other plugin has been found on the system. The
```
shamsasari, line 36: This fact should really be moved up where we mention that the Azure plugin is bundled

docs/docs/faq.md
```diff
@@ -187,7 +181,7 @@ You *don't* need to trust:
 * The hardware manufacturers *other* than the CPU vendor. RAM, disk, firmware, the operating system, PCI devices, and
 other hardware components are all untrusted in the SGX threat model.
 * The owner of the hardware on which your enclave is running. You can verify the remote attestation to ensure that the
-remote system is up to date and that the enclave is running securely.
+remote system is up-to-date and that the enclave is running securely.
```
shamsasari, line 184: I've see other instances of "up to date" in dcap.md

docs/docs/running-hello-world.md
```diff
@@ -151,9 +151,13 @@ In the previous section, you run the sample application in [mock mode](enclave-m
 suggests, isn't secure at all.
 
 This section describes how to build and run the sample in other [enclave modes](enclave-modes.md), namely
-simulation, debug, and release. Note that only release mode provides confidential guarantees of SGX.
+simulation, debug, and release.
 
-*First, ensure that your system fulfills the [system requirements](enclave-modes.md#system-requirements).*
+Only the release mode provides confidential guarantees of SGX. You can learn to develop enclaves on Windows, macOS,
```
shamsasari, line 156: The release mode statement doesn't belong with the rest of the paragrah. In fact, perhaps it should be its our info box.

shamsasari (changes requested):
@bonifazi-r3 please also take a look.

docs/docs/dcap.md
```diff
@@ -19,19 +9,33 @@ automated for you.
 As cloud providers run caches, DCAP supports vendor-specific plugins. Intel provides a default one
 which requires a [subscription](https://api.portal.trustedservices.intel.com/provisioning-certification).
 
-Azure provides a DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) that does not require a subscription. Conclave
-bundles and uses Azure's DCAP plugin by default. The Azure caches are open to the public internet and can be used from anywhere. Azure Confidential Computing instances like DC4s_v3 come pre-provisioned for DCAP and as Conclave
+Azure provides a DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) that does not require a subscription. 
+Conclave bundles and uses Azure's DCAP plugin by default. The Azure caches are open to the public internet and can 
+be used from anywhere. Azure Confidential Computing instances like DC4s_v3 come pre-provisioned for DCAP and as Conclave
```
bonifazi-r3, line 14: Here I feel there should be a `,` before `as Conclave`

docs/docs/dcap.md
bonifazi-r3, line 26: I am wondering if it makes sense to go to a newline before `Intel provides...`. So it would be more visually clearer that there is a generic DCAP client from Intel and then one from MIcrosoft Azure.

docs/docs/dcap.md
bonifazi-r3: I agree.
In fact the previous sentence `Conclave bundles and uses Azure's DCAP plugin by default` is somehow misleading. If there is another library, the other library might be considered the "default".
So, as Shams suggested, I would refactor such a sentence saying something like
`Conclave bundles the Azure's DCAP plugin, this will be used only if no other plugin has been found on the system.`. Or similar

docs/docs/dcap.md
```diff
@@ -54,11 +58,12 @@ You may want to set the Azure DCAP client logging level to FATAL as the default
 export AZDCAP_DEBUG_LOG_LEVEL=FATAL
 ```
 
-If you are using the bundled version, you can skip the rest of this section. Otherwise, read on for instructions on how to manually configure the DCAP plugin.
+If you are using the bundled version, you can skip the rest of this section. Otherwise, read on for instructions on
```
bonifazi-r3, line 61: I realize this is weird somehow actually.
Why should the user manually configure the DCAP plugin if we provide a bundled version? I feel something is odd, why were these instructions written actually @shamsasari ? Should we actually take a look and see what happens if we DO NOT install the Azure DCAP client?

bonifazi-r3: I.e. we might actually scrap the installation instructions below, which is quite unnecessary. We might just put a warning about the fact that the Intel libraries should not be there, or otherwise they would override our bundled libs

bonifazi-r3 (changes requested):

docs/docs/dcap.md
JohnJoser3: Instead, I have split the sentence into two for clarity.

docs/docs/running-hello-world.md
JohnJoser3: Instead, I have removed the sentence. It's better to avoid info boxes on this first tutorial to avoid affecting the flow. We already mention on the enclave-modes page that only the release mode is confidential. We also link to it from this tutorial. I believe that is sufficient.

docs/docs/faq.md
JohnJoser3: I've changed it back to 'up to date'. IntelliJ marked this as an error and made me think it was incorrect. Here, it is used as an adverb and should not be hyphenated. Thank you. :)

docs/docs/dcap.md
JohnJoser3: Hi @shamsasari, can you please check this?

docs/docs/dcap.md
```diff
@@ -45,20 +30,43 @@ them if necessary).
...
 ```sh
 export AZDCAP_DEBUG_LOG_LEVEL=FATAL
 ```
 
-If you are using the bundled version, you can skip the rest of this section. Otherwise, read on for instructions on how to manually configure the DCAP plugin.
+The Azure caches are open to the public internet and can be used from anywhere. Azure Confidential Computing 
+instances like DC4s_v3 come pre-provisioned for DCAP. As Conclave comes with the necessary libraries bundled, you 
+don't need to do any further setup.
+
+### DCAP Plugin
```
shamsasari, line 47: This heading is wrong. I think it should be `Provisioning Certificate Caching Service (PCCS)` or something similar.

shamsasari: I guess what I'm asking is what information is this section trying to convey? How to use PCCS, or something else?

docs/docs/dcap.md
shamsasari: @bonifazi-r3 you're assuming that the Azure plugin will work in all scenarios, including bare metal outside of Azure's Cloud?

bonifazi-r3: I am confused by the instructions even before these changes, also considering that these were written for Graal runtime.
`Otherwise, read on for instructions`.
What is `Otherwise`? I.e. in which case the user code does not actually use the Azure bundle version? The only case I can think of is when the user actually does NOT have any Azure DCAP libraries in `/usr/lib` but the Intel ones. I.e. when the user intentionally does not want to use the Azure DCAP, but they want to use the Intel's one. So in this "Otherwise" paragraph, why are we suggesting the user how to install manually the Azure DCAP client, in this hacky way?

So, my follow up question is: when would the user want to use the Azure `.deb` library from Ubuntu repo instead of using the Azure bundled one? My answer is: **never**, and we should not give further instructions for that.
We have also already clarified that the default libraries are in `/usr/lib` etc, so if we have a hacker user, this would know what to do.

And finally, why would the user need to do that weird manual process, instead of an easy `apt-get install` or similar, depending on the platform?

bonifazi-r3: Apologies for the confusion, @JohnJoser3, feel free to setup another chat to discuss.
I would personally do the following: 

- Have an introductory paragraph in relation to DCAP protocol where we do not mentions plugins (basically leave what we have now)
- Then a title like "DCAP client" and describe that it is basically a library `libdcap_quoteprov.so.1` with a symbolic link at `libdcap_quoteprov.so`  that is going to be installed in  in `/usr/lib/` or `/usr/lib/x86_64-linux-gnu` (here I would put something like "apart from the bundled one, see below") and that the users need to be careful to avoid conflicts of plugin and uninstall the ones that they do not want.

Then three subtitles:
 one for the "Intel DCAP client", another for "Azure DCAP client" and the last one one with "Conclave-Azure bundled client".

For the Intel DCAP client (in case of bare metal machines) we need to suggest how to install the package (`sudo apt-get install libsgx-dcap-default-qpl`) but also tell the users that they need to install the PCCS service and just link them the instructions.

For the Azure DCAP client we need to suggest how to install the package, for example with Ubuntu `sudo apt-get install az-dcap-client` and tell the users that they need to be on Azure machines.

As a last subtitle, I would mention that Conclave has a recommended bundled client (an Azure bundled) which is only picked up correctly if no other plugins have been installed in the default directories `/usr/lib/`.
We can, of course, reuse the wording we already have, I feel it is more matter of restructuring and remove the "manual" instructions (the ones that suggest to download and unpack the .deb file), which I don't think are needed.

I am happy to rewrite a proposal so that you can correct, or just let me know how you want to proceed

JohnJoser3: Hi @bonifazi-r3 Marco, I have rewritten the doc according to the instructions. Please take another look.

I have retained the last two sections - `Using Docker container(s)` and `Running a Conclave Application` as these sections have some additional info. Please let me know if I need to remove those too.

docs/docs/dcap.md
```diff
@@ -1,96 +1,70 @@
...
-As cloud providers run caches, DCAP supports vendor-specific plugins. Intel provides a default one
-which requires a [subscription](https://api.portal.trustedservices.intel.com/provisioning-certification).
+1. Intel DCAP client
+2. Azure DCAP client
+3. Conclave-Azure bundled client
 
-Azure provides a DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) that does not require a subscription. Conclave
-bundles and uses Azure's DCAP plugin by default. The Azure caches are open to the public internet and can be used from anywhere. Azure Confidential Computing instances like DC4s_v3 come pre-provisioned for DCAP and as Conclave
-comes with the necessary libraries bundled, you don't need to do any further setup.
+To avoid conflicts between DCAP client plugins, you must uninstall the plugins that you don't require.
```
bonifazi-r3, line 23: This is the first time we use the word `plugin` in this doc. I would keep things consistent and redundant and use client instead.
I would say something like
`To avoid conflicts between DCAP clients, you must uninstall the ones you don't require`

Or in the sentence above, we should clarify "A DCAP client (also called **DCAP plugin**)" is used for this function`

docs/docs/dcap.md
```diff
@@ -1,96 +1,70 @@
...
-The runtime will use the first `.so` it encounters in the search order below:
-```
-/usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so.1
-/usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so
-/usr/lib/libdcap_quoteprov.so.1
-/usr/lib/libdcap_quoteprov.so
-```
-Should you decide to use a bundled version (recommended), ensure the files listed above don't exist (delete or rename
-them if necessary).
+`sudo apt-get install libsgx-dcap-default-qpl`
```
bonifazi-r3, line 31: These are Ubuntu specific instructions only, should we mention that? I.e. `for example, in Ubuntu:`...

docs/docs/dcap.md
```diff
@@ -1,96 +1,70 @@
...
+### Intel DCAP Client
 
-DCAP is designed to work on many different server topologies. So, rather than directly connecting to Intel services to retrieve this information, the cloud vendor or the owner of the SGX system must provide a DCAP client plugin with the required information. Intel provides a generic DCAP client plugin as part of the DCAP runtime. To use this, you also need to set up a Provisioning Certificate Caching Service (PCCS). Intel provides an example and some instructions [here](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/pccs/README.md). If you would like to use Intel's reference implementation of their [PCCS service](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/pccs), then additional work may be required to provide data in the correct format for Conclave. Please contact R3 support if you need to set this up.
+Intel provides a DCAP client plugin as part of the DCAP runtime.
```
bonifazi-r3, line 27: Here I would make it clear why the users might want to use, i.e. then they need to build your own "DCAP infrustructure" and don't want to rely on cloud providers.
I would clarify like this a bit
```
Intel provides a DCAP client plugin as part of the DCAP runtime; you generally use this when you don't want to use cloud providers and you want to run your code on bare-metal machines.
To do that, you need to install the Intel DCAP client, for example in Ubuntu you can run this...

```

docs/docs/dcap.md
```diff
@@ -1,96 +1,70 @@
...
-    [follow the instructions below](#azure-client-plugin) to build/install an updated version of the Azure client 
-    plugin that supports the platform.
+### Azure DCAP Client
 
-You may want to set the Azure DCAP client logging level to FATAL as the default setting is quite verbose:
-```sh
-export AZDCAP_DEBUG_LOG_LEVEL=FATAL
-```
+Microsoft provides an Azure DCAP client, which you can use on Azure virtual machines. Azure's DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) 
+does not require a subscription.
```
bonifazi-r3, line 43: Here I would clarify that it is an Intel related subscription

docs/docs/dcap.md
```diff
@@ -1,96 +1,70 @@
...
 
-You may want to set the Azure DCAP client logging level to FATAL as the default setting is quite verbose:
-```sh
-export AZDCAP_DEBUG_LOG_LEVEL=FATAL
-```
+Microsoft provides an Azure DCAP client, which you can use on Azure virtual machines. Azure's DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) 
+does not require a subscription. 
 
-If you are using the bundled version, you can skip the rest of this section. Otherwise, read on for instructions on how to manually configure the DCAP plugin.
+You can use the following command to install Azure's DCAP client package:
```
bonifazi-r3, line 45: Again, should we clarify that this is just for Ubuntu?

docs/docs/dcap.md
```diff
@@ -1,96 +1,70 @@
...
+`sudo apt-get install az-dcap-client`
 
-* Identify the currently installed DCAP client plugin. It will always have one of the following names: `libdcap_quoteprov.so.1` or `libdcap_quoteprov.so`. You might find other similarly named files, but they won't be used as a plugin.
-```sh
-ls /usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so*
-ls /usr/lib/libdcap_quoteprov.so*
-```
+### Conclave-Azure Bundled Client
+
+Conclave provides a bundled Azure DCAP plugin, which is the recommended option for users. This plugin will be used
```
bonifazi-r3, line 51: This is the recommended option for users using Azure virtual machine, can we make it explicit?

bonifazi-r3 (changes requested):
I really like this now, very good work.
I have some minor comments, feel free to ping me for discussions

JohnJoser3:
I've made the changes. PTAL.

bonifazi-r3 (approved):

bonifazi-r3:
This is fine to me. @shamsasari  you requested changes

docs/docs/dcap.md
```diff
@@ -1,96 +1,73 @@
...
-```
+Microsoft provides an Azure DCAP client, which you can use on Azure virtual machines. Azure's DCAP [plugin](https://github.com/microsoft/Azure-DCAP-Client) 
+does not require any subscription to Intel services. 
 
-If you are using the bundled version, you can skip the rest of this section. Otherwise, read on for instructions on how to manually configure the DCAP plugin.
+To use an Azure DCAP client, you need to install Azure's DCAP client package. For example, in Ubuntu, you can use 
+the following command:
 
-#### Azure client plugin
+`sudo apt-get install az-dcap-client`
```
shamsasari, line 50: I don't believe this will work just on its own. The user will need to first install the Microsoft packages list, or otherwise it won't recognise `az-dcap-client`.

```
wget -qO - https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
echo 'deb [arch=amd64] https://packages.microsoft.com/ubuntu/20.04/prod focal main' | sudo tee /etc/apt/sources.list.d/msprod.list
```

docs/docs/dcap.md
```diff
@@ -1,96 +1,73 @@
...
-In DCAP, repeated attestation requests aren't forwarded to Intel, but rather served from a cache. A newly installed
-machine obtains a machine certificate from Intel via the cache, which may then be persisted to disk. All this is
-automated for you.
+You can use one of the three available DCAP clients listed below:
 
-As cloud providers run caches, DCAP supports vendor-specific plugins. Intel provides a default one
-which requires a [subscription](https://api.portal.trustedservices.intel.com/provisioning-certification).
+1. Intel DCAP client
+2. Azure DCAP client
+3. Conclave-Azure bundled client
```
shamsasari, line 21: I feel we need to better highlight that this is the default, and that in most cases you don't need to do further setup DCAP since we bundle the library for you. Not sure how best to present that. I don't want readers thinking that the need to to _do_ to get DCAP attestation working, esp if they are using Azure.

JohnJoser3: That is exactly what I did. Please take another look, @shamsasari.

shamsasari (approved):
Thanks @JohnJoser3 and @bonifazi-r3 for taking the time to get this right.

PR #111, opened 2022-12-02
CON-1178: Fix Gramine hardware integration tests by reintroducing AESM service

bonifazi-r3:
In the context of our work on Gramine attestation, we need to reintroduce the AESM service, which was previously removed.
Gramine is using the  AESM service in background to get the signed quote from the quoting enclave, and therefore we need the AESM service to run the Gramine integration tests, as they are currently hanging and timing out.

In this PR we also remove the installation of the `az-dcap-client` Azure DCAP plugin package, as we rely on the version which is bundled together with Conclave.


docs/docs/dcap.md
```diff
@@ -87,7 +87,7 @@ Please read the installation instructions in the "Install the DCAP packages" sec
 If you plan to use a Docker container with DCAP hardware, you must map two different device files like this:
 
 ```sh
-docker run --device /dev/sgx/enclave --device /dev/sgx/provision ...
+docker run --device /dev/sgx_enclave --device /dev/sgx_provision ...
```
bonifazi-r3, line 90: `/dev/sgx/enclave` is related to the old out of kernel driver, which we do not support in Conclave 1.4

containers/aesmd/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN DEBIAN_FRONTEND=noninteractive  apt-get install -y \
+    wget \
+    gnupg2
+
+RUN wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | apt-key add -
+RUN echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu focal main' > /etc/apt/sources.list.d/intel-sgx.list
+
+RUN apt-get update
+
+# TODO: Update Conclave to Intel SGX SDK 2.18 and remove package versions: https://r3-cev.atlassian.net/browse/CON-1243
```
shamsasari, line 19: Can you mention this point in that JIRA ticket so taht it's not lost on whoever picks it up.

containers/aesmd/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    sgx-aesm-service=2.17.100.3-focal1 \
```
shamsasari, line 21: In theory we should only need this package, right? Can you try removing the others and see if it still works.

containers/integration-tests-build/Dockerfile
```diff
@@ -10,16 +10,31 @@ RUN apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
...
+
 
 # Disable interactive dialogue
 # Otherwise, a message that requires user input might appear during the installation
+# TODO: Update Conclave to Intel SGX SDK 2.18 and remove package versions: https://r3-cev.atlassian.net/browse/CON-1243
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
-        az-dcap-client \
-        libsgx-quote-ex \
-        libsgx-dcap-ql
+    sgx-aesm-service=2.17.100.3-focal1 \
```
shamsasari, line 24: If we have the aesm deamon in the other container, why does it need to be installed here?

shamsasari (changes requested):

containers/aesmd/Dockerfile
bonifazi-r3: I think I got issues with dependencies. I.e. if a package has dependencies, you need to specify the version for the dependencies too.
Same for all the libsgx* ones. We only need one or two, in theory.
 I will double check again.

containers/integration-tests-build/Dockerfile
bonifazi-r3: Uh, this slipped. I forgot to remove these after I realized that I need a (root) container for the AESM. Will remove

containers/aesmd/Dockerfile
shamsasari: OK, in that case, can you comment on which packages are actually dependencies on the aesm service, and thus can be removed once we move onto the latest version.

bonifazi-r3: I confirm that dependencies' versions are not following the "established" dependent version. 
I.e. if I run this
```
sudo apt-get install libsgx-quote-ex=2.17.100.3-focal1 libsgx-dcap-ql=1.14.100.3-focal1
```
here is what I get
```
...
Setting up libsgx-ae-qe3 (1.15.100.3-focal1) ...
Setting up libsgx-dcap-quote-verify (1.15.100.3-focal1) ...
Setting up libsgx-ae-le (2.18.100.3-focal1) ...
Setting up libsgx-quote-ex (2.17.100.3-focal1) ...
Setting up libsgx-ae-id-enclave (1.15.100.3-focal1) ...
Setting up libsgx-launch (2.18.100.3-focal1) ...
Setting up libsgx-ae-qve (1.15.100.3-focal1) ...
Setting up libsgx-enclave-common (2.18.100.3-focal1) ...
Setting up libsgx-ae-epid (2.18.100.3-focal1) ...
Setting up libsgx-ae-pce (2.18.100.3-focal1) ...
Setting up libsgx-aesm-launch-plugin (2.18.100.3-focal1) ...
Setting up libsgx-urts (2.18.100.3-focal1) ...
Setting up libsgx-pce-logic (1.15.100.3-focal1) ...
Setting up libsgx-qe3-logic (1.15.100.3-focal1) ...
Setting up libsgx-dcap-ql (1.14.100.3-focal1) ...
Setting up libsgx-aesm-pce-plugin (2.18.100.3-focal1) ...
Setting up libsgx-aesm-epid-plugin (2.18.100.3-focal1) ...
Setting up libsgx-aesm-ecdsa-plugin (2.18.100.3-focal1) ...
Setting up libsgx-aesm-quote-ex-plugin (2.18.100.3-focal1) ...

```
From these I just removed the epid ones.
I was already bitten by some libraries being on 2.18 and AESM not working, so I think it is correct to set all those dependencies as I have done.

I suggest to put a TODO asking to replace all of these with what was there before, i.e. the default instructions (as suggested by Intel SGX SDK, "Intel® SGX Application User")
https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_SW_Installation_Guide_for_Linux.pdf

```
sudo apt-get install libsgx-epid libsgx-quote-ex libsgx-dcap-ql
```
of course, without `libsgx-epid`

containers/aesmd/Dockerfile
bonifazi-r3: Done

bonifazi-r3: https://r3-cev.atlassian.net/browse/CON-1243

containers/aesmd/Dockerfile
shamsasari: ok, add the TODO

PR #113, opened 2022-12-05
Added Jep into NOTICE file

shamsasari:


bonifazi-r3:
Should we add TOMLJ as well? And should I mention Pytorch example repo in my PR as well?

shamsasari:
I wasn't sure about tomlj, since it's only used in integration tests. But I'll add it in anyway.

bonifazi-r3 (approved):

PR #114, opened 2022-12-05
CON-1200: Refactor Python Enclave test by replacing it with a Pytorch example

bonifazi-r3:


bonifazi-r3:
@shamsasari I have refactored this test with an example I have taken from Pytorch. With this, the we only add a file of 11MB. The result is obviously not accurate as I had to change Pytorch parameters to make things run as fast as possible (bash size to 5000 and only 1 epoch). 
Is something like this what you had in mind?

integration-tests/python-enclave/host/src/test/kotlin/com/r3/conclave/integrationtests/python/enclave/PythonEnclaveTest.kt
```diff
@@ -39,21 +45,31 @@ class PythonEnclaveTest {
...
         enclaveHost.deliverMail(mailRequest, null)
         val mailResponse = postedMail.single().encryptedBytes
-        assertThat(clientPostOffice.decryptMail(mailResponse).bodyAsBytes).isEqualTo("54321".toByteArray())
+        // We do not really care of the accuracy of the result, we just want the Pytorch enclave to run correctly.
+        val lastLossResult = clientPostOffice.decryptMail(mailResponse).bodyAsBytes
+        assertThat(lastLossResult).isEqualTo("2.2989193603515625".toByteArray())
+    }
+
+    private fun prepareMailWithDataBundle(clientPostOffice: PostOffice): ByteArray {
+        val pytorchModel = Paths.get(this::class.java.getResource("/bundle.zip")!!.file)
```
shamsasari, line 59: TIL. I didn't realise you could get a test resouce without having to copy it.

integration-tests/python-enclave/host/src/test/kotlin/com/r3/conclave/integrationtests/python/enclave/PythonEnclaveTest.kt
```diff
@@ -39,21 +45,31 @@ class PythonEnclaveTest {
...
-        val signature = enclaveHost.callEnclave(data)!!
-        with(enclaveHost.enclaveInstanceInfo.verifier()) {
-            update(data)
-            assertThat(verify(signature)).isTrue
-        }
-    }
-
-    @Test
-    fun receive_enclave_mail() {
+    fun run_mnist_rnn_pytorch_example() {
```
shamsasari, line 48: We're no longer in Python! Use backticks and full sentence for the test method anem

shamsasari (changes requested):
Yep, this is exactly what I had in mind, thanks. Just some minor changes. Also, perhaps we should rename this module to pytorch-enclave to better represent the test?

shamsasari:
It's a shame TC didn't run the test, but I think that's an existing issue. Not all the tests run, even though `failFast` is off.

shamsasari (approved):
Thanks

PR #115, opened 2022-12-06
CON-1178: Fix certificate issues of AESM service

bonifazi-r3:
With the previous changes we have introduced the AESM service. This works but it is currently throwing an exception related to missing certificates. The root cause of the problem is that AESM service needs the Azure DCAP Client in order to retrieve such certificates and therefore we need to add it to the related AESM container.
This PR adds that.

In order to make an integration test pass (`MockEnclaveEnvironmentHardwareCompatibilityTest`), we also have to increase the `maxThread` value of the enclave used in that test

### DIFF
--- a/integration-tests/general/threadsafe-enclave/build.gradle
+++ b/integration-tests/general/threadsafe-enclave/build.gradle
@@ -11,7 +11,7 @@ conclave {
     productID = 1
     revocationLevel = 0
     runtime = runtimeType
-    maxThreads = 10         // To match assumptions in tests
+    maxThreads = 40         // To match assumptions in tests
 
     kds {
         kdsEnclaveConstraint = "S:B4CDF6F4FA5B484FCA82292CE340FF305AA294F19382178BEA759E30E7DCFE2D PROD:1 SEC:INSECURE"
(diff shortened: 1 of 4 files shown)


bonifazi-r3:
See log here https://r3-conclave.teamcity.com/buildConfiguration/ConclaveCore_SdkIntegrationTestsDebugGramine/14368?buildTab=log&focusLine=0&logView=flowAware&linesState=133
The test is now failing "genuinely" with another issue (we could suppress the test at some point).

integration-tests/general/threadsafe-enclave/build.gradle
```diff
@@ -11,7 +11,7 @@ conclave {
     productID = 1
     revocationLevel = 0
     runtime = runtimeType
-    maxThreads = 10         // To match assumptions in tests
+    maxThreads = 40         // To match assumptions in tests
```
shamsasari, line 14: We now default to 100, so perhaps we can just delete this?

bonifazi-r3: This was added by @ac101m yesterday, with the intention to be explicit I think (as this enclave is also used to test the max number of TCs). So I would still make this intention clear somehow and not rely on the default value of the SDK?

shamsasari: Fair enough.

ac101m: Marco is correct, some of the tests were assuming the default value. Best to be explicit.

shamsasari (approved):

ac101m (commented):
LGTM, will leave it to @shamsasari for final approval

PR #101, opened 2022-11-21
CON-1203: Gradle plugin to use conclave-build container on Linux

anien:

PR #125, opened 2022-12-16
CON-1203: Non-Linux envs must always use Docker for building the enclave

shamsasari:
https://github.com/R3Conclave/conclave-core-sdk/pull/101 accidently made building on MacOS and Windows conditional on the new `buildInDocker` flag. This is not correct as Docker is always required for these environments.

### DIFF
--- a/plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
+++ b/plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
@@ -6,12 +6,18 @@ import org.gradle.api.GradleException
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
+import org.gradle.internal.os.OperatingSystem
 import java.io.ByteArrayOutputStream
 import java.io.File
-import java.io.IOException
 import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import javax.inject.Inject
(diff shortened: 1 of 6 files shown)


shamsasari:
Since you have an Intel MBP can you also test this @anien. Thanks

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -52,58 +54,69 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
...
-                )
+                val message = if (OperatingSystem.current().isLinux) {
+                    "Conclave requires Docker to be installed when building enclaves. Please install Docker and " +
+                            "rerun your build. See https://docs.conclave.net/enclave-modes.html#system-requirements " +
+                            "for more information. If the build still fails, please rerun the build with the " +
+                            "--stacktrace flag and raise an issue at https://github.com/R3Conclave/conclave-core-sdk/issues/new"
+                } else {
+                    "Conclave requires Docker to be installed when building enclaves on non-Linux platforms. Please " +
+                            "install Docker and rerun your build. See " +
+                            "https://docs.conclave.net/tutorial.html#setting-up-your-machine and " +
```
anien, line 65: https://docs.conclave.net/tutorial.html#setting-up-your-machine returns 404.

shamsasari:
PTAL

anien (approved):
Looks good!

I have tested this on macOS and it works fine. Scenario: add `buildInDocker = false` flag to `enclave/build.gradle`, build simulation mode, run host and client - works. Build debug mode - works.

shamsasari:
Thanks Anna!

PR #126, opened 2022-12-16
CON-1232: Build and run Gramine enclaves inside Docker

bonifazi-r3:
Allow Gramine enclaves to be built and run inside a Docker container.


bonifazi-r3:
@shamsasari can you give me an early feedback on this? This makes Gramine running in the `conclave-build` container

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
```diff
@@ -11,5 +11,6 @@ object PluginUtils {
     const val GRAMINE_SGX_TOKEN = "java.token"
     /** This contains the enclave's [SgxEnclaveCss], also known as the `SIGSTRUCT`. */
     const val GRAMINE_SIGSTRUCT = "java.sig"
+    const val GRAMINE_SECCOMP = "gramine-seccomp.json"
```
bonifazi-r3, line 14: This file will need to be passed in the bundle, for the same reasons of the integration tests @shamsasari you fixed.

conclave-host/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     implementation "io.github.classgraph:classgraph:$classgraph_version"
     // Logging using the SLF4J API
     implementation "org.slf4j:slf4j-api:$slf4j_version"
+    implementation "org.tomlj:tomlj:$tomlj_version"
```
bonifazi-r3, line 16: I need TOML to read the tag that I passed in the Gramine manifest and that needs to be  used by the host

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -52,8 +60,27 @@ class GramineEnclaveHandle(
...
         enclaveInterface = SocketHostEnclaveInterface()
+
+        dockerImageTag = retrieveImageTagFromManifest()
+    }
+
+    private fun String.mapWorkingDirectory(): String {
+        return this.replace(workingDirectory.toFile().absolutePath, DOCKER_WORKING_DIR).replace("\\", "/")
+    }
+
+    private fun runSimpleCommand(command: List<String>): String {
```
bonifazi-r3, line 71: This is needed to run `id -u`. Couldn't find a better way, feel free to suggest

containers/conclave-build/Dockerfile
```diff
@@ -11,23 +11,24 @@ ENV LC_ALL en_US.UTF-8
...
+# Disable interactive dialogue
+# Otherwise, a message that requires user input will appear while installing Java 17
 RUN apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
         build-essential \
         bash \
         curl \
         apt-transport-https \
-        python3-pip
+        python3-pip \
+        openjdk-17-jdk-headless
```
bonifazi-r3, line 22: Not sure why this was actually after adding the Gramine repository, I feel it is nicer to separate the installation after adding the Gramine repository.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -132,13 +158,14 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
                 "as the native image build process is memory intensive."
     )
 
-    private fun getDockerRunArgs(params: List<String>): List<String> {
+    private fun getDockerRunArgs(params: List<String>, dockerWorkdirPath: String?): List<String> {
```
bonifazi-r3, line 161: Not sure if this mechanism of passing `dockerWorkdirPath` can be improved or even simplified. Will double check on Monday

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/gramine-seccomp.json
```diff
@@ -0,0 +1,781 @@
+{
```
shamsasari, line 1: Is this the same file as the one in the root of the repo? If so then just use that and copy it into the conclave-host.jar. There's no need to put it in the gramine bundle.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -175,14 +175,17 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
...
     ): GenerateGramineBundle {
-        return target.createTask("generateGramine${enclaveMode.capitalise()}Bundle", enclaveMode) { task ->
+        return target.createTask("generateGramine${enclaveMode.capitalise()}Bundle", enclaveMode, linuxExec) { task ->
             // TODO: Build Gramine enclaves in conclave-build container: https://r3-cev.atlassian.net/browse/CON-1229
+            task.dependsOn(linuxExec)
             task.signingKey.set(signingKey)
             task.productId.set(conclaveExtension.productID)
             task.revocationLevel.set(conclaveExtension.revocationLevel)
             task.maxThreads.set(conclaveExtension.maxThreads)
+            task.dockerImageTag.set(linuxExec.tag)
```
shamsasari, line 188: The tag to use must be the hash that we calculate over the containers directory. It can't be the CONCLAVE_SDK_VERSION as it's not specific enough.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -98,32 +98,58 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
...
+            }
         }
-        result.assertNormalExitValue()
-        return null
     }
 
-    /** Returns the output of the command executed in the container. */
-    fun execWithOutput(params: List<String>): String {
-        return commandWithOutput(*getDockerRunArgs(params).toTypedArray())
+    fun execWithOutput(params: List<String>, dockerWorkdirPath: String? = null, throwsException: Boolean=false): String {
```
shamsasari, line 135: We seem to be duplicating a lot of command execution code. Why does this exist?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -52,8 +59,27 @@ class GramineEnclaveHandle(
...
+        val process = ProcessBuilder()
+            .command(command)
+            .start()
+        process.waitFor(5L, TimeUnit.SECONDS)
+        val lines = process.inputStream.bufferedReader().use(BufferedReader::readText).trimEnd()
+            .split(System.lineSeparator())
+        //  We expect 1 line result and a newline
+        check(lines.size == 1) { "Command produced an unexpected result" }
```
shamsasari, line 78: If this is only needed to get the `id` then just return the whole string. no need to parse it into lines and then assert there's only one.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -62,14 +88,9 @@ class GramineEnclaveHandle(
...
-        val command = mutableListOf(getGramineExecutable(enclaveMode))
-        command += listOf(
-            "java",
-            "-cp",
-            GRAMINE_ENCLAVE_JAR,
-            "com.r3.conclave.enclave.internal.GramineEntryPoint",
-            port.toString()
-        )
+        val user = runSimpleCommand(listOf("id", "-u"))
+        val group = runSimpleCommand(listOf("id", "-g"))
```
shamsasari, line 92: Use varargs

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
shamsasari: See comment below, but I don't seewhy we need to copy this file from the enclave plugin to the host. If it's needed by both the plugin and host, then put it in `conclave-common`.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -137,6 +158,57 @@ class GramineEnclaveHandle(
...
+            dockerImageTag
+        )
+    }
+
+    private fun retrieveImageTagFromManifest(): String {
+        val manifest = Paths.get(
+            if (enclaveMode == EnclaveMode.SIMULATION) "$workingDirectory/$GRAMINE_MANIFEST" else "$workingDirectory/$GRAMINE_SGX_MANIFEST"
+        )
+        val parseResult = Toml.parse(manifest)
+        return parseResult.getString("host.env.CONCLAVE_DOCKER_IMAGE_TAG")!!
```
shamsasari, line 209: The conclave-build tag should be pass into the jar manifest via gradle. @filipesoliveira has some ideas. I suggest the tag be written into the conclave-common.jar since it's needed by both the plugin and the host.

containers/conclave-build/Dockerfile
shamsasari: Move the setting of JAVA_HOME next to this.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
```diff
@@ -94,6 +94,15 @@ abstract class ConclaveTask : DefaultTask() {
         }
     }
 
+    fun commandWithResultAndOutput(
```
shamsasari, line 97: I'm actually not sure we need these command helpers. The Gradle API has the `project.exec` feature. If you need the exec result then perhaps just use that directly rather than creating another method.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
```diff
@@ -94,6 +94,15 @@ abstract class ConclaveTask : DefaultTask() {
...
+        command: List<String>,
+        commandLineConfig: CommandLineConfig
+    ): Pair<ExecResult, String> {
+        val cmd = command.joinToString(" ")
```
shamsasari, line 101: This isn't used

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -62,28 +69,30 @@ open class GenerateGramineBundle @Inject constructor(
     val outputDir: DirectoryProperty = objects.directoryProperty()
 
     override fun action() {
-        enclaveJar.copyToOutputDir(GRAMINE_ENCLAVE_JAR)
+
```
shamsasari, line 72: delete

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -93,63 +102,77 @@ open class GenerateGramineBundle @Inject constructor(
...
     }
 
+    private fun copySeccompToOutputDir() {
+        val seccomp = outputDir.get().asFile.resolve(GRAMINE_SECCOMP).toPath()
+        javaClass.copyResource(GRAMINE_SECCOMP, seccomp)
+    }
+
+    private fun retrieveArchitecture(): String {
+        return linuxExec.execWithOutput(listOf("gcc", "-dumpmachine"))
+    }
```
shamsasari, line 112: No need to call this as we know what the architecture is inside the container

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -62,28 +69,30 @@ open class GenerateGramineBundle @Inject constructor(
...
+        val architecture = retrieveArchitecture()
         val ldPreload = executePython(
             "from sysconfig import get_config_var; " +
                     "print(get_config_var('LIBPL') + '/' + get_config_var('LDLIBRARY'))"
         )
         // The location displayed by 'pip3 show jep' is actually of the site/dist-packages dir, not the specific 'jep'
         // dir within it. We assume this is the packages dir for other modules as well. If this assumption is
         // incorrect then we'll need to come up with a better solution.
-        val pythonPackagesPath = commandWithOutput("pip3", "show", "jep")
+        val pythonPackagesPath = linuxExec.execWithOutput(listOf("pip3", "show", "jep"))
```
shamsasari, line 88: Leave python as is. we don't want it executing inside the container in this PR

shamsasari (changes requested):

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
bonifazi-r3: Yes, technically this is needed only by the host, so can embed this there.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
bonifazi-r3: Yes, this is a temporary solution to make things running on my side. But still don't have clear, should be us to create the "hash" to identify the image? In this PR? If so, I suggest to use the list of package versions installed  (`dpkg`) + our files to do that, or is there any other thought on this?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: Ok, just wanted to prevent potential failures. I noticed that we used `LinuxExec.exec` without checking the (null) return value. This has issues when I used Docker, i.e. failures in running commands inside Docker weren't propagated and no exceptions were thrown

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
bonifazi-r3: I think we need to refactor the "normal" LinuxExec.exec`. This does not give the possibility of getting back the output when the command succeeds and does not throw exceptions when failing.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: Ok, but what we do with it then? I.e. should we remove Python related code? Should we leave the current flow without running things inside Docker?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: Ok, so you mean the `conclave-common.jar` manifest then. The plugin would write in the `conclave-common.jar` manifest and then read it on the host side, right? Isn't a bit weird anyway to leave this information in a generic file, i.e. I see it nice to use the Gramine manifest as a mechanism also to pass all Gramine related host and enclave info. 
Not a real issue though.
Regarding the tag, as I asked above: is the generation of this hash tag part of this PR then? I'd use the list of packages + fat jar + manifest files to create it, if so.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
shamsasari: The tag is already calculated. It's the hash of the containers dir, and we do it here:
https://github.com/R3Conclave/conclave-core-sdk/blob/f26ba0b649cb663649568d12a70ad61169383d40/scripts/ci_build_common.sh#L29
The trick is to get this value and put it into the jar manifest. @filipesoliveira has the same task for his [ticket](https://r3-cev.atlassian.net/browse/CON-1069), so I suggest you two collaborate on the best solution.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
shamsasari: Yes, leave things as they are for Python and not have it execute inside the container. It's captured in a separate [ticket](https://r3-cev.atlassian.net/browse/CON-1229), which may or may not be done as part of 1.4.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
shamsasari: >The plugin would write in the conclave-common.jar manifest and then read it on the host side, right? 
No, our building of the SDK, i.e. conclave-common/build.gradle, will need to get the tag and insert it into the jar manifest. Since both the host and the plugin use conclave-common, they should then have access to this tag value.

>Isn't a bit weird anyway to leave this information in a generic file, i.e. I see it nice to use the Gramine manifest as a mechanism also to pass all Gramine related host and enclave info.
I disagree. This build tag is a Conclave implementation detail and not strictly a Gramine concern. Also, it requires parsing the TOML file and the extra dependency. Using the manifest avoids that.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
filipesoliveira: @bonifazi-r3, we can discuss the solution today. I have found a way to call scripts from Gradle and capture the output to a variable inside Gradle.

containers/conclave-build/Dockerfile
bonifazi-r3: I also moved the CMD comand closer

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -63,27 +66,39 @@ open class GenerateGramineBundle @Inject constructor(
 
     override fun action() {
         enclaveJar.copyToOutputDir(GRAMINE_ENCLAVE_JAR)
+        // Building enclaves with Python has been temporarily disabled
+        // TODO: CON-1215 - Building enclaves with Python inside a Docker container
+        check(!pythonFile.isPresent) { "Building enclaves with Python has been disabled" }
```
bonifazi-r3, line 71: @shamsasari I still don't have clear what I need to do about Python now: my understanding is that I should not keep the Gramine-Non-docker flow, so should I just leave Python unusable with a TODO as it is here? Or even remove code that will be then readded (this sounds a bit weird).

shamsasari: No, Python should still work, but not use the conclave-build container and instead rely on all the components installed on the local machine. You should still be able to run the PyTorch demo with your changes: https://github.com/R3Conclave/conclave-samples/tree/master/pytorch

bonifazi-r3: Does this mean that we will actually keep the current way of building enclaves then, just for Python?

shamsasari: Correct. I assume this is not a problem, and just a matter of conditional statement.

conclave-common/build.gradle
```diff
@@ -18,3 +18,11 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertj_version"
     testImplementation "org.junit.jupiter:junit-jupiter:$junit_jupiter_version"
 }
+
+jar {
+    manifest {
+        attributes("Conclave-Version": rootProject.version.toString(),
+                   "Jep-Version": jep_version.toString(),
```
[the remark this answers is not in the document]
shamsasari: No, let's not do that. Only add the manifest attributes where they're needed. Jep is just a concern of the plugin and so should go there.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -79,7 +84,7 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
     // TODO Come up with a better way than this. This might not be an issue after CON-1069 since we won't be building
     //  the conclave-build image anymore.
     fun buildInDocker(buildInDocker: Property<Boolean>): Boolean {
-        return !OperatingSystem.current().isLinux || buildInDocker.get()
+        return !OperatingSystem.current().isLinux || buildInDocker.get() || runtimeType.get() == GRAMINE
```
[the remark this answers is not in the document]
shamsasari: Building in `conclave-build` should be the default. Only for GraalVM is the `buildInDocker` flag which can turn this off. The reason this flag exists is purely to make sure we haven't accidently prevented existing users have built enclaves in 1.3 from building in 1.4 bc of a Docker issue we didn't encounter. For Gramine this isn't a concern since it's a new feature of 1.4 and thus there are not any existing users to potentially disturb.

Your change here looks correct.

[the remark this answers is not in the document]
shamsasari: Also, for Python the `conclave-build` isn't used, but that's just a prioritisation issue. If we have time before 1.4 to get that working inside `conclave-build` then great, but we don't want to risk 1.4 bc of it.

conclave-common/build.gradle
```diff
@@ -18,3 +18,11 @@ dependencies {
...
+                   "Conclave-Build-Image-Tag": "conclave-build:$conclave_version")
```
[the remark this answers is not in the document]
shamsasari: Could @filipesoliveira's work be merged first? Or can you not simply copy the relevant bit from this branch?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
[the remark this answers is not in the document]
bonifazi-r3: Ok. I made this question because I had previously understood that we should not allow building enclaves (Java and Python) outside the container.
I would keep Python out of the build in the container and revert some changes so that we have conditional code to have `conclave-build` Java enclaves and normal out-of-container build for Python.

conclave-common/build.gradle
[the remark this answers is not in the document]
bonifazi-r3: Ok, great

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: Ok, great, will readd back the conditional code

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: ok

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
bonifazi-r3: Moved to the host side

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: We now call this only when using Python

conclave-common/build.gradle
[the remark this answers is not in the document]
bonifazi-r3: Kept these separate, I guess I will refactor them again depending on Filipe's changes

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -30,30 +31,16 @@ import org.gradle.util.VersionNumber
...
-            return when (values.size) {
-                1 -> values.first()
-                0 -> throw IllegalStateException("Could not find manifest attribute $name")
-                else -> throw IllegalStateException("Found multiple values for manifest attribute $name: $values")
-            }
-        }
+        private val CONCLAVE_SDK_VERSION =
+            getManifestAttribute(this::class.java.classLoader, "Conclave-Release-Version")
+        private val CONCLAVE_GRAALVM_VERSION =
+            getManifestAttribute(this::class.java.classLoader, "Conclave-GraalVM-Version")
```
bonifazi-r3, line 43: I'd like to save to have another `getManifestAttribute` hence I am passing the classLoader. Is that ok?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/NativeImage.kt
```diff
@@ -429,48 +429,50 @@ open class NativeImage @Inject constructor(
...
-        if (errorOut?.any { "Image generator watchdog is aborting image generation" in it } == true) {
-            // If there is too much memory pressure in the container, native image can just get slower and slower until
-            // a watchdog timer fires. Give the user some advice on how to fix it.
-            linuxExec.throwOutOfMemoryException()
-        }
-        else if (errorOut?.any { "Default native-compiler executable 'gcc' not found via environment variable PATH" in it } == true) {
-            // This error will only occur on Linux because on other platforms native-image is run in a docker container that
-            // already has gcc installed.
-            throw GradleException(
+        } catch (e: GradleException) {
```
bonifazi-r3, line 459: Not sure why before we were relying on the return of `LinuxExec.exec` to handle the errors. I feel it is more correct to throw the exception anyway (avoiding swallow). And now that I actually need the output, I think it is nicer to catch and rethrow in case of exceptions.

bonifazi-r3:
@shamsasari Would you mind to take another look, while we decide when to link our PRs?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -137,6 +157,59 @@ class GramineEnclaveHandle(
...
+            "java",
+            "-cp",
+            GRAMINE_ENCLAVE_JAR,
+            "com.r3.conclave.enclave.internal.GramineEntryPoint",
+            port.toString()
+        )
+    }
+
+    private fun getDockerCommand(user: String, group: String): List<String> {
+        javaClass.copyResource("/$GRAMINE_SECCOMP", workingDirectory.toAbsolutePath() / GRAMINE_SECCOMP)
```
shamsasari, line 187: `.toAbsolutePath()` not needed here

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -137,6 +157,59 @@ class GramineEnclaveHandle(
...
+        val workingDirectoryPath = workingDirectory.toFile().absolutePath
```
shamsasari, line 188: `workingDirectory.absolutePathString()`

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -137,6 +157,59 @@ class GramineEnclaveHandle(
...
+            "run",
+            "-u",
+            "$user:$group",
+            "--network",
+            "host",
+            "--device=/dev/sgx_enclave",
+            "--device=/dev/sgx_provision",
+            "-v",
+            "/var/run/aesmd:/var/run/aesmd",
+            "-i",
```
shamsasari, line 200: Is this needed?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -137,6 +157,59 @@ class GramineEnclaveHandle(
...
+            "docker",
+            "run",
+            "-u",
+            "$user:$group",
+            "--network",
+            "host",
+            "--device=/dev/sgx_enclave",
+            "--device=/dev/sgx_provision",
+            "-v",
+            "/var/run/aesmd:/var/run/aesmd",
```
shamsasari, line 199: Can you put these two together on the same line to make it clearer they are needed together

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -137,6 +157,59 @@ class GramineEnclaveHandle(
...
+            "-i",
+            "--rm",
+            //  Map the host's working directory to the Docker container's working directory
+            "-v",
+            "$workingDirectoryPath:$DOCKER_WORKING_DIR",
```
shamsasari, line 204: ditto

conclave-host/src/main/resources/gramine-seccomp.json
```diff
@@ -0,0 +1,781 @@
+{
```
shamsasari, line 1: How is this file different to the one in `scripts` directory?

containers/conclave-build/Dockerfile
shamsasari: I don't think the java -version cmd is necesary

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
shamsasari: The classloader doesn't matter in this scan. You can use the `PluginUtils::class.java.classLoader` in the method and it will give the same results.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -235,12 +224,6 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
         return (project.properties["sourceSets"] as SourceSetContainer).getByName("main")
     }
 
-    private fun createMockArtifact(target: Project, enclaveFatJar: Jar) {
-        // Mock mode does not require all the enclave building tasks. The enclave Jar file is just packaged
-        // as an artifact.
-        target.artifacts.add("mock", enclaveFatJar.archiveFile)
-    }
-
```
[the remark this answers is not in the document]
bonifazi-r3: Where is this used? https://github.com/R3Conclave/conclave-core-sdk/search?q=createMockArtifact

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -23,11 +23,11 @@ sgx.isvsvn    = {{ isv_svn }}
 {% if is_python_enclave == "true" %}
 loader.env.LD_LIBRARY_PATH = "/lib:{{ arch_libdir }}:/usr{{ arch_libdir }}:/usr/lib:{{ python_packages_path }}/jep"
 loader.env.PYTHONPATH = "{{ python_packages_path }}"
+loader.env.LD_PRELOAD = "{{ python_ld_preload }}"
```
shamsasari, line 26: I would keep the arg just `ld_preload`. We might need it even without Python.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
```diff
@@ -9,7 +13,23 @@ object PluginUtils {
     const val GRAMINE_MANIFEST = "java.manifest"
     const val GRAMINE_SGX_MANIFEST = "java.manifest.sgx"
     const val GRAMINE_SGX_TOKEN = "java.token"
+    const val GRAMINE_SECCOMP = "gramine-seccomp.json"
```
shamsasari, line 16: This is not used by the plugin I don't think.

conclave-host/src/main/resources/gramine-seccomp.json
bonifazi-r3: No difference. The one in `scripts` is used for integration tests though and somehow I feel like it is part of a separate project.
Not sure how we can reuse that nicely?

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
```diff
@@ -9,7 +13,23 @@ object PluginUtils {
...
+    const val DOCKER_WORKING_DIR = "/project"
```
shamsasari, line 17: There's no reason why the mounted dir when running the enclave needs to be the same as the mounted directory when the enclave is built. But this is a minor point since we'll be trying to move away from docker at runtime.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
```diff
@@ -44,10 +48,10 @@ open class ConclaveExtension @Inject constructor(objects: ObjectFactory) {
     // Constants for the two types we support. Allows the user to not have to use string quotes if they don't want to.
     @Suppress("unused")
     @get:Internal
-    val graalvm = "graalvm"
+    val graalvm = GRAALVM
     @Suppress("unused")
     @get:Internal
-    val gramine = "gramine"
+    val gramine = GRAMINE
```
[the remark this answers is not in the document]
bonifazi-r3: Ok, but that has not changed I think. My intention was to have a const to wrap the string. I reuse the `GRAMINE` const here
` runtimeType.get() == GRAMINE` in the condition to run in docker

conclave-host/src/main/resources/gramine-seccomp.json
shamsasari: Something like
```
processResources {
  from "${rootProject.projectDir}/scripts/gramine-seccomp.json"
}
```

in conclave-host/build.gradle.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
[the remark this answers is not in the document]
shamsasari: Ah, use `runtimeType` in `GradleEnclavePlugin`, as it parses the string to the enum.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: Good idea, will do the same for `-u` where I used it

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
bonifazi-r3: I agree, but it was convenient I think. I can put a comment actually. Also the `enclave.jar` needs actually to be on the same path both when building and running, but we can leverage the Docker workdir parameter for that.

conclave-common/src/main/kotlin/com/r3/conclave/common/internal/PluginUtils.kt
bonifazi-r3: Correct, it was a leftover. I will move this in `GramineEnclaveHandle` were it is actually used.

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
bonifazi-r3: I am ok to rename it to `ld_preload`, but I would keep this still inside the Python presence condition.

I was bitten by an error then I felt it is clearer to change it/use it only when actually needed.

My understanding is also that we currently rely on Python being installed in order to feed this `LD_PRELOAD` parameter that is not useful for non-Python enclaves. This is somehow not nice/incorrect.
But, more importantly, providing an empty value would cause issues when Gramine generates the manifest.
So I felt that moving this inside the python presence condition was better.

bonifazi-r3: Actually I can try to check for the emptiness in jinja, will try that

bonifazi-r3: Ok, reverted and changed, seems fine, have a look

conclave-host/src/main/resources/gramine-seccomp.json
bonifazi-r3: I would have actually done the opposite, i.e. reusing the host resource in the integration test.
But yes, I will got with this solution, thanks!

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: Removed. And things are running fine.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
bonifazi-r3: Fine, is this because they are in the same package right? I find a bit weird to add the import for that, will use `PluginUtils` as you suggest anyway

containers/conclave-build/Dockerfile
bonifazi-r3: Removing

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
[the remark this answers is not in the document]
bonifazi-r3: This should be done now

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -26,6 +31,10 @@ class GramineEnclaveHandle(
 
     companion object {
         private val logger = loggerFor<GramineEnclaveHandle>()
+        private const val GRAMINE_SECCOMP = "gramine-seccomp.json"
+
+        private val dockerImageTag =
+            getManifestAttribute(PluginUtils::class.java.classLoader, "Conclave-Build-Image-Tag")
```
shamsasari, line 37: You don't need to pass in the classloader. Have `getManifestAttribute` use it directly.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -54,6 +63,16 @@ class GramineEnclaveHandle(
         enclaveInterface = SocketHostEnclaveInterface()
     }
 
+    private fun runSimpleCommand(vararg command: String): String {
+        val process = ProcessBuilder()
+            .command(command.asList())
+            .start()
+        process.waitFor(5L, TimeUnit.SECONDS)
+        return process.inputStream.bufferedReader().use(BufferedReader::readText).trimEnd()
+            .removeSuffix(System.lineSeparator())
```
shamsasari, line 72: There's some redundancy here and over-complication. Just do `process.inputStream.reader().use { it.readText() }.trimEnd()`

containers/conclave-build/Dockerfile
```diff
@@ -11,26 +11,25 @@ ENV LC_ALL en_US.UTF-8
...
 #  command and then removed.
+# Disable interactive dialogue
+# Otherwise, a message that requires user input will appear while installing Java 17
 RUN apt-get -y update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
         build-essential \
         bash \
         curl \
         apt-transport-https \
-        python3-pip
+        python3-pip \
```
shamsasari, line 21: Ah, just realised. Can we remove Python from `conclave-build` for now, since we're not using `conclave-build` for building Python enclaves. We'll see when we do CON-1229 if Python should be in `conclave-build` or a new image that extends it.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -183,10 +171,12 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
             enclaveMode: EnclaveMode,
             conclaveExtension: ConclaveExtension,
             enclaveFatJarTask: Jar,
-            signingKey: Provider<RegularFile?>
+            signingKey: Provider<RegularFile?>,
+            linuxExec: LinuxExec
     ): GenerateGramineBundle {
-        return target.createTask("generateGramine${enclaveMode.capitalise()}Bundle", enclaveMode) { task ->
+        return target.createTask("generateGramine${enclaveMode.capitalise()}Bundle", enclaveMode, linuxExec) { task ->
             // TODO: Build Gramine enclaves in conclave-build container: https://r3-cev.atlassian.net/browse/CON-1229
```
shamsasari, line 178: Delete!

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -79,7 +83,7 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
     // TODO Come up with a better way than this. This might not be an issue after CON-1069 since we won't be building
     //  the conclave-build image anymore.
     fun buildInDocker(buildInDocker: Property<Boolean>): Boolean {
-        return !OperatingSystem.current().isLinux || buildInDocker.get()
+        return !OperatingSystem.current().isLinux || buildInDocker.get() || runtimeType.get() == GradleEnclavePlugin.RuntimeType.GRAMINE
```
shamsasari, line 86: Just add a comment that Gramine enclaves will always be built in the image, whilst GraalVM enclaves will be by default but can be opted out with the `buildInDocker` config.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -128,10 +137,11 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
             errorOut.writeTo(System.err)
             // Using default charset here because the strings come from a sub-process and that's what they'll pick up.
             // Hopefully it's UTF-8 - it should be!
-            return String(errorOut.toByteArray()).split(System.lineSeparator())
+            val errorMessage =  String(errorOut.toByteArray())
+            throw GradleException(errorMessage)
         }
         result.assertNormalExitValue()
-        return null
+        return String(stdOut.toByteArray()).split(System.lineSeparator())
```
shamsasari, line 144: We shouldn't really split the output here, it's the caller's job to do that if they wish. It's only bc of `LinuxExec` we do this. I think the splitting should go there, but it's not really important for this PR.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -128,10 +137,11 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
             errorOut.writeTo(System.err)
             // Using default charset here because the strings come from a sub-process and that's what they'll pick up.
             // Hopefully it's UTF-8 - it should be!
-            return String(errorOut.toByteArray()).split(System.lineSeparator())
+            val errorMessage =  String(errorOut.toByteArray())
```
shamsasari, line 140: `errorOut.toString()`

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/NativeImage.kt
shamsasari: I agree

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/NativeImage.kt
```diff
@@ -429,48 +429,50 @@ open class NativeImage @Inject constructor(
...
                     "Conclave requires gcc to be installed when building GraalVM native-image based enclaves. "
                             + "Try running 'sudo apt-get install build-essential' or the equivalent command on your distribution to install gcc. "
                             + "See https://docs.conclave.net/tutorial.html#setting-up-your-machine"
-            )
-        }
-        else if (errorOut != null) {
-            throw GradleException("The native-image enclave build failed. See the error message above for details.")
+                )
+            } else {
+                throw GradleException("The native-image enclave build failed. See the error message above for details.")
```
shamsasari, line 474: Just throw `e` as is.

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
```diff
@@ -69,7 +71,6 @@ sgx.trusted_files = [
     "file:{{ arch_libdir }}/",
     "file:/usr/{{ arch_libdir }}/",
     "file:{{ java_home }}/",
-    "file:{{ ld_preload }}",
```
shamsasari, line 72: If `ld_preload` is specified then we do need this in the trusted files list.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -63,15 +66,33 @@ open class GenerateGramineBundle @Inject constructor(
...
         enclaveJar.copyToOutputDir(GRAMINE_ENCLAVE_JAR)
+
+        val manifestTemplatePath = temporaryDir.resolve(MANIFEST_TEMPLATE).toPath()
+        javaClass.copyResource(MANIFEST_TEMPLATE, manifestTemplatePath)
+
         if (pythonFile.isPresent) {
-            pythonFile.copyToOutputDir(PYTHON_FILE)
+            generateManifestForPythonEnclaves(manifestTemplatePath.absolutePathString())
+        } else {
+            generateManifestForJavaEnclaves(manifestTemplatePath.absolutePathString())
```
shamsasari, line 76: Please use `Path` as much as possible, only converting to strings where necessary.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -63,15 +66,33 @@ open class GenerateGramineBundle @Inject constructor(
...
+        }
+    }
+
+    private fun generateManifestForPythonEnclaves(manifestTemplatePath: String) {
+        // We currently build Python enclaves outside of the conclave-build container
+        // TODO: CON-1215 - Building enclaves with Python inside a Docker container
+        pythonFile.copyToOutputDir(PYTHON_FILE)
+
         // TODO We're relying on gcc, python3, pip3 and jep being installed on the machine that builds the Python
         //  enclave. https://r3-cev.atlassian.net/browse/CON-1181
```
shamsasari, line 93: I think we can delete this TODO.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -82,74 +103,79 @@ open class GenerateGramineBundle @Inject constructor(
...
-        val manifestTemplateFile = temporaryDir.resolve(MANIFEST_TEMPLATE).toPath()
-        javaClass.copyResource(MANIFEST_TEMPLATE, manifestTemplateFile)
+    private fun generateManifestForJavaEnclaves(manifestTemplatePath: String) {
+        val command = prepareManifestGenerationCommand(
+            DOCKER_IMAGE_ARCHITECTURE,
+            "",
+            "",
+            manifestTemplatePath
+        )
+        linuxExec.exec(command, listOf("-w", outputDir.get().asFile.absolutePath))
```
shamsasari, line 126: Why not just use `execCommand`? Yes, we know that it's not Python here, but it avoids some duplication and eventually Python will be executed inside the container.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -82,74 +103,79 @@ open class GenerateGramineBundle @Inject constructor(
...
+        )
 
-        generateManifest(architecture, ldPreload, pythonPackagesPath)
-
-        if (enclaveMode != EnclaveMode.SIMULATION) {
-            generateSgxManifestAndSigstruct()
-            generateToken()
-            // The .manifest is not needed for debug and release modes
-            outputDir.file(GRAMINE_MANIFEST).get().asFile.toPath().deleteExisting()
+        project.exec { spec ->
```
shamsasari, line 113: Again, use `execCommand`

shamsasari:
This is looking good now. Let's get https://github.com/R3Conclave/conclave-core-sdk/pull/132 approved and merged first.

containers/conclave-build/Dockerfile
bonifazi-r3: Yes, let's remove this specific package.
Note that Gramine will install Python, it's a dependency (Gramine scripts are Python scripts) 
```
marco@bonifazi:~$ sudo apt-cache depends gramine
gramine
  Depends: libcjson1
  Depends: libcurl4
  Depends: libprotobuf-c1
  Depends: python3
  Depends: python3-click
  Depends: python3-cryptography
  Depends: python3-jinja2
  Depends: python3-protobuf
  Depends: python3-pyelftools
  Depends: python3-toml
  Conflicts: gramine-dcap
  Conflicts: gramine-oot
  Recommends: gramine-ratls-dcap
  Recommends: gramine-ratls-epid
```

plugin-enclave-gradle/src/main/resources/com/r3/conclave/plugin/enclave/gradle/gramine/java.manifest.template
bonifazi-r3: Good one

bonifazi-r3: Not really sure actually why I tried to remove it, I guess because it was only meant for Python. Will readd back with the if condition.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: Yes, I did exactly because Python enclaves are not yet in. Fine to use it, sure!

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
bonifazi-r3: Added

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
shamsasari: Yes, more or less.

bonifazi-r3: Sorry, I realize what you mean now. In Filipe's PR we have the attributes all in the same module. Will do the same, now I can actually merge, so things are easier

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: When this is used I actually need this as a string. Therefore I preferred to pass around the actual think I needed. Not a big deal anyway, reverted to be a path and then converting at the end

conclave-common/build.gradle
```diff
@@ -18,3 +18,24 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertj_version"
     testImplementation "org.junit.jupiter:junit-jupiter:$junit_jupiter_version"
 }
+
+jar {
+    manifest {
+        attributes(
+                "Conclave-GraalVM-Version": conclave_graal_version.toString(),
```
shamsasari, line 25: This is only needed by the plugin, so for better clarity it should be set in there.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -137,6 +156,55 @@ class GramineEnclaveHandle(
...
+            "--device=/dev/sgx_enclave",
+            "--device=/dev/sgx_provision",
+            "-v", "/var/run/aesmd:/var/run/aesmd",
+            "--rm",
+            //  Map the host's working directory to the Docker container's working directory
+            "-v", "$workingDirectoryPath:$DOCKER_WORKING_DIR",
+            //  Set the directory internally used in Docker as the working directory
+            "-w", DOCKER_WORKING_DIR,
+            "--security-opt",
+            "seccomp=$workingDirectoryPath/$GRAMINE_SECCOMP",
```
shamsasari, line 203: same line

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
```diff
@@ -94,9 +94,9 @@ abstract class ConclaveTask : DefaultTask() {
         }
     }
 
-    fun commandWithOutput(vararg command: Any): String {
+    fun commandWithOutput(vararg command: String, workingDir: String? = null): String {
```
shamsasari, line 97: I don't think the varags works well with a parameter after it.

shamsasari, line 97: workingDir should be a `Path`.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
```diff
@@ -94,9 +94,9 @@ abstract class ConclaveTask : DefaultTask() {
...
         val output = ByteArrayOutputStream()
-        commandLine(command.asList(), CommandLineConfig(standardOutputStream = output))
+        commandLine(command.asList(), CommandLineConfig(standardOutputStream = output, workingDir = workingDir ))
```
shamsasari, line 99: trailing space

shamsasari (approved):
This looks good. Happy to approve and get this merged. The remaining comments can be addressed in your next PR.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
bonifazi-r3: I think it that it works if you specify the names of the parameters (when you use both)

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: Done

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
bonifazi-r3: I'll leave this to another PR then. Maybe it is also a good idea to change the parameter of the other functions called in the call stack

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveTask.kt
bonifazi-r3: done

conclave-common/build.gradle
bonifazi-r3: As discussed offline, I have added a comment (and a TODO).
If we do not pass the classloader to the `getManifestAttribute` function, only the manifest of the classloader module will be accessible.
We could readd the classloader parameter of the manifest or find other solution

shamsasari:
Can you enable `PytorchEnclaveTest` that was disabled in https://github.com/R3Conclave/conclave-core-sdk/pull/134. Your changes here should be enough to fix it.

integration-tests/general/filesystem-db-enclave/build.gradle
```diff
@@ -38,16 +38,16 @@ conclave {
 
     simulation {
         signingType = privateKey
-        signingKey = file("../signing/sample_private_key.pem")
+        signingKey = file("sample_private_key.pem")
```
shamsasari, line 41: Why did this need to be done?

integration-tests/general/threadsafe-enclave-same-signer/build.gradle
```diff
@@ -13,18 +13,21 @@ conclave {
     runtime = runtimeType
     useInternalDockerRegistry = true
 
+    //  Note that the content of thread_safe_enclave_private_key_same_signer.pem is the same as
+    //    thread_safe_enclave_private_key.pem in thread_safe_enclave module.
+    //  We avoid relative paths here as Gramine tests would fail otherwise
```
shamsasari, line 18: Do we have a ticket for this? I presume this will affect users as well. What was the error?

integration-tests/pytorch-enclave/host/src/test/kotlin/com/r3/conclave/integrationtests/python/enclave/PytorchEnclaveTest.kt
```diff
@@ -54,7 +52,8 @@ class PytorchEnclaveTest {
         val mailResponse = postedMail.single().encryptedBytes
         // We do not really care of the accuracy of the result, we just want the Pytorch enclave to run correctly.
         val lastLossResult = clientPostOffice.decryptMail(mailResponse).bodyAsBytes
-        assertThat(lastLossResult).isEqualTo("2.2989193603515625".toByteArray())
+        val diff = kotlin.math.abs(String(lastLossResult).toDouble() - 2.2989193603515625)
+        assertThat(diff).isLessThanOrEqualTo(0.0001)
```
shamsasari, line 56: I believe assertJ has an `.closeTo` method which takes in a margin or error with doubles.

integration-tests/general/threadsafe-enclave-same-signer/build.gradle
bonifazi-r3: I think it is because when we build the enclave in Docker we are in a mounted directory and the "above" directories are not available. I will reproduce the issue.
But do we really want to fix it? I.e. can't we just tell the users to have the signing key inside the enclave dirs?

PR #127, opened 2022-12-19
CON-1258: Do not include graal-sdk dependency for Gramine enclaves

shamsasari:


plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -39,10 +39,13 @@ import kotlin.io.path.*
...
     companion object {
-        private const val CONCLAVE_GRAALVM_VERSION = "22.0.0.2-1.4-SNAPSHOT"
-
         private val CONCLAVE_SDK_VERSION = getManifestAttribute("Conclave-Version")
 
+        // We manually specify the Conclave version, rather than appending CONCLAVE_SDK_VERSION, because it might be
+        // possible for the two versions to go out of sync. For example, Conclave SDK is 1.4-SNAPSHOT, but it's still
+        // using 20.0.0.2-1.3 because we've not had the need to update GraalVM since 1.3.
+        private const val CONCLAVE_GRAALVM_VERSION = "22.0.0.2-1.4-SNAPSHOT"
+
```
ac101m, line 48: This comment is confusing, it took me a minute or two to parse.
Maybe rephrase to something like:

```
// Rather than appending the Conclave SDK version here, we hard code it here so the SDK and
// version conclave graal version can go out of sync.
```

I also don't think the example is really necessary.

P.S. Are we not publishing a new graal build for each version of conclave? If that's the case, why do we format the version string this way? Maybe we should be doing something like `<graal-version>-<patch-version>` where `patch-version` is a version number independent of the SDK version string.

conclave-enclave/build.gradle
```diff
@@ -8,8 +8,12 @@ dependencies {
 
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
     compileOnly "com.fasterxml.jackson.core:jackson-databind:$jackson_version"
+    // The GraalVM SDK is not needed if the user chooses the Gramine runtime, so we don't want to include it
+    // unnecessarily. It's declared as "compileOnly" to make sure everything still compiles, but the
+    // actual dependency is only added by the plugin if the user choose GraalVM as the runtime. Only then will it be
+    // included in the enclave fat jar.
```
ac101m, line 14: Comment is a bit verbose, maybe slim it down a little?

```
// The GraalVM SDK is only needed when the graalvm runtime is selected. We add it here as a
// compile time dependency to ensure compilation is successful. It is added to the enclave fat jar 
// later if required.
```

ac101m (commented):
A couple minor comments

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
shamsasari: It's certainly a valid alternative. The nice thing with this is it gives some context of which version of the SDK it's meant for, or rather the first version of the SDK.

ac101m (approved):
LGTM

PR #128, opened 2022-12-19
Remove unused files

bonifazi-r3:
In this small PR we remove some files that we realized are not currently used.


shamsasari (approved):
Thanks

PR #131, opened 2022-12-20
Set input arguments to lowercase in integration tests script

anien:
Setup parameters to lowercase so they can be properly checked in the script to spin AESM container.


shamsasari (commented):
Can I ask for something better than this, and that's to convert it all to lowercase, and then update our scripts to refer to the strings in lowercase?

Your change also won't work if they pass in all uppercase, for example.

anien:
I still want to test this first and check if we need to do something similar in the other scripts.

scripts/ci_integration_tests.sh
```diff
@@ -4,10 +4,11 @@ set -xeuo pipefail
 script_dir=$(dirname ${BASH_SOURCE[0]})
 source ${script_dir}/ci_build_common.sh
 
-enclaveMode=$1
-runtimeType=$2
+# Set parameters to lowercase
+enclaveMode=${1,,}
+runtimeType=${2,,}
```
shamsasari, line 9: To the best way to make sure this works is check the values are valid.

anien: Do we want to check the values are valid in the script? It's only TC and us using it, so I don't expect having completely wrong input. But don't mind adding the check if required.

shamsasari: Well, @filipesoliveira had to make some TC changes in the last week or so, so I think it's worth it if it's straightforward to do.

anien: Done, PTAL.

scripts/ci_integration_tests.sh
```diff
@@ -4,10 +4,20 @@ set -xeuo pipefail
...
+    exit 1
+fi
+
+# Set parameters to lowercase
+enclaveMode=${1,,}
+runtimeType=${2,,}
+
+# Check the entered enclave mode and runtime type
+[[ "mock simulation debug release" =~ (^|[[:space:]])$enclaveMode($|[[:space:]]) ]] && echo $enclaveMode || ( echo "Wrong enclave mode entered: $enclaveMode."; exit 1 )
+[[ "graalvm gramine" =~ (^|[[:space:]])$runtimeType($|[[:space:]]) ]] && echo runtimeType || ( echo "Wrong runtime type entered: $runtimeType."; exit 1 )
```
shamsasari, line 18: Wow!

shamsasari (approved):

PR #132, opened 2022-12-20
CON-1069 - Update Gradle plugin to pull conclave-build image 

filipesoliveira:
By default, the end user builds its own `conclave-build`.

The configuration property `useInternalDockerRepo` was added to the Conclave's Gradle configuration block. This property allows the integration tests to pull the `conclave-build` docker image. By default, the value of the new property is false.

Updated the integration tests to fetch the docker container `conclave-build`.


docs/docs/running-hello-world.md
```diff
@@ -225,10 +225,11 @@ To generate `host-release.jar`:
...
+=== "macOS/Windows"
 
-    You need a Linux environment to run the host in simulation mode. This section describes how to create such an
-    environment using Conclave's Docker integration.
+    Conclave needs a Linux environment to run the host in simulation mode. But you don't need to do anything as long as
+    Docker is installed on your system. You can run the host and client as we did for mock mode. The only differences are that the host will be
+    run from inside the container, and the host JAR will be named `host-simulation.jar` rather than `host-mock.jar`.
 
     !!!Note
         * You need Docker to run the *host* on macOS. You can run the *client* on macOS without Docker.
```
shamsasari, line 235: "macOS" referenced when the section is now either mac or windows.

integration-tests/general/default-enclave/build.gradle
```diff
@@ -15,4 +15,5 @@ conclave {
     productID = 1
     revocationLevel = 0
     runtime = runtimeType
+    useInternalDockerRepo = true
```
shamsasari, line 18: I just learnt that the correct term is "registry", so can you rename this to `useInternalDockerRegistry`.

plugin-enclave-gradle/build.gradle
```diff
@@ -158,11 +158,22 @@ processResources {
     }
 }
 
+def genDockerContainerTag() {
+    new ByteArrayOutputStream().withStream { os ->
+        def result = exec {
```
bonifazi-r3, line 163: Is `result` used here?

plugin-enclave-gradle/build.gradle
```diff
@@ -158,11 +158,22 @@ processResources {
...
+            executable = 'bash'
+            args "-c", "../scripts/genDockerImageTag.sh" // N.B.: The script must only print to the standard output the docker image tag. Printing other information will cause issues
+            standardOutput = os
+        }
+        return os.toString()
```
shamsasari, line 168: Also add `.trim()` incase there are any trailing lines, etc.

plugin-enclave-gradle/build.gradle
```diff
@@ -158,11 +158,22 @@ processResources {
     }
 }
 
+def genDockerContainerTag() {
+    new ByteArrayOutputStream().withStream { os ->
+        def result = exec {
+            executable = 'bash'
+            args "-c", "../scripts/genDockerImageTag.sh" // N.B.: The script must only print to the standard output the docker image tag. Printing other information will cause issues
```
bonifazi-r3, line 165: Can we indent better this comment? It goes a bit beyond the right margin

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/ConclaveExtension.kt
```diff
@@ -30,7 +30,7 @@ open class ConclaveExtension @Inject constructor(objects: ObjectFactory) {
     @get:Input
     val deadlockTimeout: Property<Int> = objects.property(Int::class.java).convention(10)
     @get:Input
-    val buildInDocker: Property<Boolean> = objects.property(Boolean::class.java).convention(true)
+    val useInternalDockerRepo: Property<Boolean> = objects.property(Boolean::class.java).convention(false)
```
shamsasari, line 33: As discussed offline, this needs to be a separate flag and which is only needed by the `LinuxExec` task. Keep `buildInDocker` as is.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -292,11 +293,11 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
         val linuxExec = target.createTask<LinuxExec>("setupLinuxExecEnvironment") { task ->
             task.dependsOn(copyGraalVM)
             task.baseDirectory.set(target.projectDir.toPath().toString())
-            task.tag.set("conclave-build:$CONCLAVE_SDK_VERSION")
+            task.tag.set("conclave-docker-dev.software.r3.com/com.r3.conclave/conclave-build:$DOCKER_CONCLAVE_BUILD_TAG")
             // Create a 'latest' tag too so users can follow our tutorial documentation using the
             // tag 'conclave-build:latest' rather than looking up the conclave version.
-            task.tagLatest.set("conclave-build:latest")
-            task.buildInDocker.set(conclaveExtension.buildInDocker)
+            task.tagLatest.set("conclave-docker-dev.software.r3.com/com.r3.conclave/conclave-build:latest")
```
shamsasari, line 299: Can we get rid of this. It's not correct as we are passing in the correct tag for the user automatically.

shamsasari (changes requested):

docs/docs/running-hello-world.md
```diff
@@ -225,10 +225,11 @@ To generate `host-release.jar`:
     On a Linux machine that meets the [system requirements](enclave-modes.md#system-requirements), you can run your
     app in all the modes the same way as in mock mode.
 
-=== "macOS"
+=== "macOS/Windows"
```
shamsasari, line 228: Please have a look at the docs changes @JohnJoser3

bonifazi-r3 (changes requested):
I don't understand the renaming of `buildInDocker`, was that one meant to be like this?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/GradleEnclavePlugin.kt
```diff
@@ -292,11 +293,9 @@ class GradleEnclavePlugin @Inject constructor(private val layout: ProjectLayout)
         val linuxExec = target.createTask<LinuxExec>("setupLinuxExecEnvironment") { task ->
             task.dependsOn(copyGraalVM)
             task.baseDirectory.set(target.projectDir.toPath().toString())
-            task.tag.set("conclave-build:$CONCLAVE_SDK_VERSION")
-            // Create a 'latest' tag too so users can follow our tutorial documentation using the
-            // tag 'conclave-build:latest' rather than looking up the conclave version.
-            task.tagLatest.set("conclave-build:latest")
+            task.tag.set("conclave-docker-dev.software.r3.com/com.r3.conclave/conclave-build:$DOCKER_CONCLAVE_BUILD_TAG")
```
shamsasari, line 296: Does `conclave-docker.software.r3.com` work?

docs/docs/running-hello-world.md
filipesoliveira: Can you point me to those changes? Does not master have the latest doc changes?

shamsasari: No, that was a comment to John!

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -28,18 +28,18 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
...
+    val useInternalDockerRegistry: Property<Boolean> = objects.property(Boolean::class.java)
 
     override fun action() {
         // This task should be set as a dependency of any task that requires executing a command in the context
         // of a Linux system or container.
         // Building the enclave requires docker container to make the experience consistent between all OSs.
         // This helps with using Gramine too, as it's included in the docker container and users don't need to
         // installed it by themselves. Only Python Gramine enclaves are built outside the container.
-        if (buildInDocker(buildInDocker)) {
+        if (buildInDocker(buildInDocker) && !useInternalDockerRegistry.get()) {
```
shamsasari, line 42: On line 56 we threw an exception with a message to notify the user that they need to install docker. I presume we don't get that same exception now when they run without building the image? Can you try testing your changes with the flag set to true but docker not installed. The user should get the same message telling them to install docker.

shamsasari (approved):
This looks good. Approving to unblock @bonifazi-r3, but would like John to take a look at the docs changes, and if Filipe you could test when docker isn't installed to make sure we get the correct error message.

After this, can you update the hello world tutorial to have the flag set to true. We need to have this tested by QA and that's the simplest way to do that.

docs/docs/running-hello-world.md
```diff
@@ -225,70 +225,19 @@ To generate `host-release.jar`:
...
 
-    You need a Linux environment to run the host in simulation mode. This section describes how to create such an
-    environment using Conclave's Docker integration.
+    Conclave needs a Linux environment to run the host in simulation mode. But you don't need to do anything as long as
+    Docker is installed on your system. You can run the host and client as we did for mock mode. The only differences are that the host will be
+    run from inside the container, and the host JAR will be named `host-simulation.jar` rather than `host-mock.jar`.
```
bonifazi-r3, line 232: `will be run from inside`. Why not `will run inside`?

docs/docs/running-hello-world.md
```diff
@@ -225,70 +225,19 @@ To generate `host-release.jar`:
     On a Linux machine that meets the [system requirements](enclave-modes.md#system-requirements), you can run your
     app in all the modes the same way as in mock mode.
 
-=== "macOS"
+=== "macOS/Windows"
 
-    You need a Linux environment to run the host in simulation mode. This section describes how to create such an
-    environment using Conclave's Docker integration.
+    Conclave needs a Linux environment to run the host in simulation mode. But you don't need to do anything as long as
```
bonifazi-r3, line 230: Not sure how clear this is and was before. 
The title is
`macOs/Windows` and the first sentence is "Conclave needs a Linux environment", which sounds odd to me.
Maybe it is not strictly relevant to this PR but I would say something more direct and clear 
```
You need a Linux Docker container to run the host in simulation mode on macOS and Windows.
```

plugin-enclave-gradle/build.gradle
```diff
@@ -158,11 +158,23 @@ processResources {
     }
 }
 
+def genDockerContainerTag() {
+    new ByteArrayOutputStream().withStream { os ->
+        exec {
+            executable = 'bash'
+            args "-c", "../scripts/genDockerImageTag.sh" // N.B.: The script must only print to the standard output
```
bonifazi-r3, line 165: Not a big deal, but can we avoid to have this relative `../` way of calling the script? I see that we used `projectDir` elsewhere, I am wondering if it is nicer to be consistent.

scripts/genDockerImageTag.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Ensure the code is run at the root directory of the project
+code_host_dir=$(dirname $(dirname ${BASH_SOURCE[0]}))
+pushd ${code_host_dir} > /dev/null # Do not print the stack to standard output otherwise it will break the plugin-enclave-gradle function getDockerContainerTag
+
+source ./containers/scripts/common.sh
```
bonifazi-r3, line 8: Not an expert of bash scripts, but do we need this? And why `./` and not `../` here, as we are in `conclave-core-sdk/scripts`?
Also, not related to this PR, do you know why we have other scripts inside the `containers/scripts` directory, i.e. why are they spread around?

bonifazi-r3 (changes requested):

scripts/genDockerImageTag.sh
```diff
@@ -0,0 +1,52 @@
...
+  # by older release branches. Keep in mind that temporary or build directories should be excluded
+  # The following code generates the hash based on the contents of a directory and the version of graal used.
+  # This hash takes into account the contents of each file inside the directory and subdirectories
+  # The cut command removes the dash at the end.
+  # All subdirectories with name build and download and hidden files are excluded. Please be sure that any file
+  # that is not tracked by git should not be included in this hash.
+  # In order to allow ci_build_publish_docker_images to detect automatically the new version of graal, the hash generated
+  # must include the conclave_graal sha512sum as well.
+  containers_dir_hash=$(find ./containers \( ! -regex '.*/\..*\|.*/build/.*\|.*/downloads/.*' \) -type f -print0 | LC_ALL=C sort -z | xargs -0 sha256sum | sha256sum | cut -d ' ' -f1)
+  docker_image_tag=$(echo $containers_dir_hash-$(getConclaveGraalSha512Sum) | sha256sum | cut -d ' ' -f1)
```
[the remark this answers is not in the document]
shamsasari: The hash tag covers the entire `containers` directory, so yes. We should make an improvement later on where only the bits relevant to `conclave-build` are used. I don't think it's a major now.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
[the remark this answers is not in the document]
shamsasari: Doesn't really matter. Most of this code will go away when we remove the flag and use docker hub by default.

docs/docs/running-hello-world.md
shamsasari: Yeah, maybe revert the merging of mac and windows. It's not really relevant to this PR and can be done by John if necessary.

plugin-enclave-gradle/build.gradle
shamsasari: `${rootProject.projectDir}`

scripts/genDockerImageTag.sh
shamsasari: Filipe knows more the details, but those scripts are needed to build the images, and since they affect the final built image, they need to be part of the hash tag.

filipesoliveira: > Also, not related to this PR, do you know why we have other scripts inside the containers/scripts directory, i.e. why are they spread around?

Some scripts change the contents of a docker container image so they have to contribute to hash generated for the docker container tag.

shamsasari (approved):
Looks good. Please check the the error message for when docker isn't installed.

docs/docs/running-hello-world.md
JohnJoser3: Please rewrite as: "The difference is that you must run the host from inside the container, and the host JAR is named `host-simulation.jar` rather than `host-mock.jar`.

docs/docs/running-hello-world.md
```diff
@@ -237,58 +238,6 @@ To generate `host-release.jar`:
...
-    3. Run the host and client as you did for mock mode. The only differences are that the host must be run from inside
-      the container, and the host JAR is named `host-simulation.jar` rather than `host-mock.jar`.
-    4. Press `CTRL+D` to exit the container when finished.
-
-    Alternatively, you can also install the Windows subsystem for Linux (2) and download Ubuntu 20.04 from the Windows
-    store. From there, you can launch an Ubuntu 20.04 shell from the start menu and build or run Conclave applications
-    as in a native Linux environment.
-    Instructions for installing WSL-2 can be found [here](https://docs.microsoft.com/en-us/windows/wsl/install).
-    Please [talk to us on Discord](https://discord.gg/zpHKkMZ8Sw) if you need help with this.
-
```
JohnJoser3, line 290: Is this section still valid?

JohnJoser3 (commented):
A few comments on the docs changes.

filipesoliveira:
> Looks good. Please check the the error message for when docker isn't installed.

I am looking into it.

docs/docs/running-hello-world.md
filipesoliveira: The end user can still use WLS  but the docs don't have to point that out because now there is no configuration needed to run Conclave on Windows. What are your thoughts, @shamsasari ?

filipesoliveira:
This PR is going to be merged because it is blocking https://github.com/R3Conclave/conclave-core-sdk/pull/126/files. A new PR will be raised if issues are uncovered while performing the remaining tests.

The remaining tests consist of checking what exceptions are raised when the user does not have docker installed.

bonifazi-r3 (approved):

PR #135, opened 2022-12-21
CON-1069 - Throw a helpful exception when docker is not installed at build time

filipesoliveira:


filipesoliveira:
Exception raised while building the host from the hello-world sample without docker installed and the flag `useInternalDockerRegistry` set to `true`:

image

shamsasari (approved):

PR #136, opened 2022-12-21
CON-1266 - Pin down Gramine version

filipesoliveira:
Updated the Dockerfile to ensure that the package manage `apt` installs a specific version of Gramine and not the latest.
Updated how the tag used for the docker images is generated by including Gramine's and Jep's versions.
Included the version of Gramine in the manifest of conclave-common.jar.


shamsasari (commented):
What happens if just the gramine version is updated? Will a new docker tag be created? I don't think it will.

filipesoliveira:
> What happens if just the gramine version is updated? Will a new docker tag be created? I don't think it will.
Was that scenario overlooked for `jep_version`?

shamsasari:
>Was that scenario overlooked for jep_version?

Yes it was. I've informed Marco of the issue.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -50,6 +51,8 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory) : ConclaveTask(
                     "--tag", tag.get(),
                     "--build-arg",
                     "jep_version=$JEP_VERSION",
+                    "--build-arg",
+                    "gramine_version=$GRAMINE_VERSION",
```
shamsasari, line 55: I think the solution is to simply hardcode the version here, since we don't need it anywhere else.

scripts/genDockerImageTag.sh
```diff
@@ -36,8 +36,9 @@ function genDockerImageTag() {
   # that is not tracked by git should not be included in this hash.
   # In order to allow ci_build_publish_docker_images to detect automatically the new version of graal, the hash generated
   # must include the conclave_graal sha512sum as well.
+  # The values for any version argument used inside the Dockerfile must also be included in the hash
   containers_dir_hash=$(find ./containers \( ! -regex '.*/\..*\|.*/build/.*\|.*/downloads/.*' \) -type f -print0 | LC_ALL=C sort -z | xargs -0 sha256sum | sha256sum | cut -d ' ' -f1)
-  docker_image_tag=$(echo $containers_dir_hash-$(getConclaveGraalSha512Sum) | sha256sum | cut -d ' ' -f1)
+  docker_image_tag=$(echo $containers_dir_hash-$(getConclaveGraalSha512Sum)-$(getJepVersion)-$(getGramineVersion) | sha256sum | cut -d ' ' -f1)
```
shamsasari, line 41: Or do this!

shamsasari: FYI @bonifazi-r3 the jep version issue has been resolved.

shamsasari (approved):
Looks good.

filipesoliveira:
> > Was that scenario overlooked for jep_version?
> 
> Yes it was. I've informed Marco of the issue.

Ok. I have covered that case too.

shamsasari:
Please wait until https://github.com/R3Conclave/conclave-core-sdk/pull/126 is merged first

PR #138, opened 2022-12-23
CON-1261: Remove usage of Docker container when running Java enclaves

bonifazi-r3:
In this patch we remove the need of using Docker to run Java enclaves.
When using Java and Gramine only few libraries are required and therefore we can narrow down the scope of the system libraries that gets used. In order to have a minimal set of Java libraries we also use JLink, which allows the creation of a customized runtime image.


filipesoliveira:
PR seems reasonable. I will let Shams approve it.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -85,6 +89,61 @@ open class GenerateGramineBundle @Inject constructor(
...
+        val enclaveJar = outputDir.file(GRAMINE_ENCLAVE_JAR).get().asFile.absolutePath
+
+        val dependentModules = execCommand(
+            "jdeps",
+            "--print-module-deps",
+            "--ignore-missing-deps",
+            enclaveJar
+        ).run { trimEnd().replace(System.lineSeparator(), "").split(",") }
+        // TODO: Provide mechanism to allow the user to add modules needed when using reflection
+        val defaultModules = listOf("java.base", "java.desktop", "java.sql", "jdk.crypto.ec")
```
shamsasari, line 102: Can you elaborate more on why these modules are needed for reflection.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -85,6 +89,61 @@ open class GenerateGramineBundle @Inject constructor(
...
+        return (defaultModules + dependentModules).distinct().joinToString(separator = ",")
+    }
+
+    private fun createCustomJDK() {
+        val jlinkOutputPath = "${outputDir.get()}/$JLINK_OUTPUT_DIRECTORY"
```
shamsasari, line 107: We should be creating a tar gz file containing the jdk image and the libs. It will preserve the permission flags on the files and enable better compression ability (since everything is compressed together in a tar file).

shamsasari (changes requested):
`conclave-build` should "extend" from `conclave-run`, then only `conclave-build` will be used to create the gramine bundle.

bonifazi-r3:
Why do we actually need two containers, considering that picking up files from `/usr/lib` seems to work?

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

containers/conclave-run/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:20.04
+
+RUN apt-get -y update
```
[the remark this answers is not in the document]
bonifazi-r3: No need. I think it is common practice, in order to pick up the most recent curl, but this works without it as well.

containers/conclave-run/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y curl
+RUN curl -fsSLo /usr/share/keyrings/gramine-keyring.gpg https://packages.gramineproject.io/gramine-keyring.gpg
+RUN echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/gramine-keyring.gpg] https://packages.gramineproject.io/ focal main' | tee /etc/apt/sources.list.d/gramine.list
+RUN DEBIAN_FRONTEND=noninteractive apt-get remove -y curl
+
+RUN apt-get -y update
+
+ARG gramine_version
+# binutils is required to use objcopy when running jlink
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y binutils openjdk-17-jdk-headless gramine=$gramine_version
```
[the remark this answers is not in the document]
bonifazi-r3: The is something I am missing. What do you mean by "runtime" here? I actually don't understand the `conclave-run` container.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -22,38 +22,53 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory, private val isP
     ConclaveTask() {
 
     companion object {
+        const val CONCLAVE_RUN_IMAGE = "conclave-run"
```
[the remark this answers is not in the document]
bonifazi-r3: Wondering why IntelliJ does NOT warn me on this.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: `java.desktop` and `java.sql` are needed by integration tests. `jdk.crypto.ec` are needed in the context of attestation (only debug and release mode affected). 
If I recollect correctly this is where it was failing
``
`fun ByteCursor<SgxEcdsa256BitPubkey>.toPublicKey(): PublicKey {
    val encodedKey = ByteBuffer.allocate(P256_HEAD.size + this.size)
    encodedKey.put(P256_HEAD)
    encodedKey.put(this.buffer)
    val keySpec = X509EncodedKeySpec(encodedKey.array())
    return KeyFactory.getInstance("EC").generatePublic(keySpec)
}
```

File: plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt (hunk 2 above)
Comment (bonifazi-r3) [reply plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt:107]: You mean that a tar gz would compress more than what is already compressed by the bundle zip? Happy to do so, I just thought it was a bit weird to zip (with zip and tar cz) twice

File: containers/conclave-run/Dockerfile (hunk 3 above)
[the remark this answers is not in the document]
Comment (shamsasari) [reply containers/conclave-run/Dockerfile:3]: Yeah maybe we can remove this then.

File: containers/conclave-run/Dockerfile (hunk 4 above)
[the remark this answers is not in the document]
Comment (shamsasari) [reply containers/conclave-run/Dockerfile:14]: As discussed offline, let's remove `conclave-run`.

File: plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt (hunk 1 above)
Comment (shamsasari) [reply plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt:102]: As discussed offline, let's add `jdk.crypto.ec` in the SDK, and add a config to allow the user to add extra modules for their enclaves.

File: plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt (hunk 2 above)
Comment (shamsasari) [reply plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt:107]: Yes, a tar gz will compress better and it will preserve the flags on the files.

File: plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt [hunk 6]
```diff
@@ -85,6 +89,61 @@ open class GenerateGramineBundle @Inject constructor(
...
 
+    private fun findJavaModules(): String {
+        val enclaveJar = outputDir.file(GRAMINE_ENCLAVE_JAR).get().asFile.absolutePath
+
+        val dependentModules = execCommand(
+            "jdeps",
+            "--print-module-deps",
+            "--ignore-missing-deps",
+            enclaveJar
+        ).run { trimEnd().replace(System.lineSeparator(), "").split(",") }
```
shamsasari, line 100: Isn't the output of `jdeps` already comma separated? There should be no need to any more processing here.

shamsasari, line 100: I don't think we need the `replace(System.lineSeparator(), "")`

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/Utils.kt
```diff
@@ -6,3 +6,8 @@ import java.nio.file.Path
 operator fun Path.div(other: String): Path = resolve(other)
 
 inline fun <reified T> ObjectFactory.newInstance(vararg parameters: Any): T = newInstance(T::class.java, *parameters)
+
+enum class RuntimeType {
```
shamsasari, line 10: This isn't really a utility. Best to have this in its own file.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
shamsasari, line 107: something like `outputDir.get().asFile.toPath().resolve(JLINK_OUTPUT_DIRECTORY)` to keep the type as a `Path`.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -85,6 +89,61 @@ open class GenerateGramineBundle @Inject constructor(
...
+            "/usr/lib/x86_64-linux-gnu/libstdc++.so.6",
+            "/usr/lib/x86_64-linux-gnu/libgcc_s.so.1"
+        )
+        val outputSystemDir = outputDir.get().asFile.toPath() / SYSTEM_LIB_DIRECTORY
+
+        if (outputSystemDir.exists()) {
+            outputSystemDir.toFile().deleteRecursively()
+        }
+        files.forEach {
+            val filePath = Paths.get(outputSystemDir.absolutePathString() + it)
```
shamsasari, line 137: `outputSystemDir.resolve(it)` should work as well.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: Do we have a nice recursive delete in Kotlin? I see that it is experimental, but that is the main reason of why I wanted to use File here

shamsasari: `.toFile().deleteRecursively()` is your best bet i think

plugin-enclave-gradle/build.gradle
```diff
@@ -71,11 +71,14 @@ processResources {
...
     String sgxSdkBuildDir = "$cppBuildDir/linux-sgx/linux-sgx-ext-prefix/src/linux-sgx-ext-build/sgxsdk"
 
-    // conclave-build Dockerfile
     into("conclave-build") {
         from("${rootProject.projectDir}/containers/conclave-build")
     }
 
+    into("conclave-run") {
+        from("${rootProject.projectDir}/containers/conclave-run")
+    }
```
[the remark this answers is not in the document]
bonifazi-r3: Removed `conclave-run`

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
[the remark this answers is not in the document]
bonifazi-r3: Removed `conclave-run`

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: It gets messed up because of the initial `/`. I would leave it as it is.

conclave-host/build.gradle
```diff
@@ -11,6 +11,8 @@ dependencies {
     compileOnly "com.fasterxml.jackson.core:jackson-databind:$jackson_version"
 
     implementation "io.github.classgraph:classgraph:$classgraph_version"
+    implementation "org.apache.commons:commons-compress:$apache_commons_compress"
```
shamsasari, line 14: We need to add this to the NOTICE file. Or use the `tar` command.

shamsasari: Sorry, ignore my comment on using `tar` command. Just update the NOTICE file

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -150,20 +158,46 @@ class GramineEnclaveHandle(
         return false
     }
 
+    private fun unTarFile(tarGz: Path, outputDir: Path) {
+        tarGz.inputStream().use { fis ->
+            GZIPInputStream(fis).use { gis ->
+                TarArchiveInputStream(gis).use { tis ->
```
shamsasari, line 164: Closing `TarArchiveInputStream` will close all the underlying streams as well. So you can simplify this to
```
TarArchiveInputStream(GZIPInputStream(tarGz.inputStream())).use { tar ->

}
```

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -150,20 +158,46 @@ class GramineEnclaveHandle(
...
+                    var tarEntry = tis.nextTarEntry
+
+                    while (tarEntry != null) {
+                        val outputFile = outputDir / tarEntry.name
+                        if (tarEntry.isDirectory) {
+                            if (!outputFile.exists()) {
+                                outputFile.toFile().mkdirs()
```
shamsasari, line 171: `outputFile.createDirectories()``

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -150,20 +158,46 @@ class GramineEnclaveHandle(
...
+    private fun unTarFile(tarGz: Path, outputDir: Path) {
+        tarGz.inputStream().use { fis ->
+            GZIPInputStream(fis).use { gis ->
+                TarArchiveInputStream(gis).use { tis ->
+                    var tarEntry = tis.nextTarEntry
+
+                    while (tarEntry != null) {
+                        val outputFile = outputDir / tarEntry.name
+                        if (tarEntry.isDirectory) {
+                            if (!outputFile.exists()) {
```
shamsasari, line 170: You don't need to do this check with `createDirectories`

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -150,20 +158,46 @@ class GramineEnclaveHandle(
...
+                                outputFile.toFile().mkdirs()
+                            }
+                        } else {
+                            outputDir.toFile().mkdirs()
+                            FileOutputStream(outputFile.toFile()).use {
+                                IOUtils.copy(tis, it)
+                            }
```
shamsasari, line 177: Use `Files.copy`

integration-tests/tribuo/enclave/build.gradle
```diff
@@ -32,6 +32,8 @@ conclave {
         signingType = dummyKey
     }
 
+    extraJavaModules = "java.logging,java.xml"
```
shamsasari, line 35: Is it easy to make this config into a list?

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -87,6 +93,80 @@ open class GenerateGramineBundle @Inject constructor(
...
+            addFilesToTarGZ(sourceDir, "", tarOs)
+        } catch (e: IOException) {
+            e.printStackTrace()
+        } finally {
+            try {
+                tarOs!!.close()
+            } catch (e: IOException) {
+                e.printStackTrace()
+            }
+        }
```
shamsasari, line 169: Delete this exception handling. If there's an IO error then let it propagate as is.

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -87,6 +93,80 @@ open class GenerateGramineBundle @Inject constructor(
...
+
+    private fun addFilesToTarGZ(filePath: String, parent: String, tarArchive: TarArchiveOutputStream) {
+        val file = File(filePath)
+        val entryName = parent + file.name
+        tarArchive.putArchiveEntry(TarArchiveEntry(file, entryName))
+        if (file.isFile) {
+            val fis = FileInputStream(file)
+            val bis = BufferedInputStream(fis)
+
+            IOUtils.copy(bis, tarArchive)
```
shamsasari, line 139: Stick with `Path` API and use `Files.copy`

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: I think that what I was trying to do here was to copy from the stream to the file, is `Files.copy` providing this?

shamsasari: Yes

integration-tests/tribuo/enclave/build.gradle
bonifazi-r3: Probably, I actually followed the same approach we use for `supportLanguages` though, should we be consistent?

containers/conclave-run/Dockerfile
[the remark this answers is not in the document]
bonifazi-r3: WE removed the container, marking solved

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: This is done

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: Sure

conclave-host/build.gradle
bonifazi-r3: Done, have a look

shamsasari (commented):
There are merge conflicts to fix. I would have made another set of comments if it weren't for the situation (unless you do want them), so happy to approve.

bonifazi-r3:
> There are merge conflicts to fix. I would have made another set of comments if it weren't for the situation (unless you do want them), so happy to approve.

Sorry, haven't noticed them, merging. There was another merge before and thought I was done.

shamsasari (commented):
Thanks Marco.

shamsasari (approved):

PR #139, opened 2022-12-28
CON-1228 Release mode Gramine enclave to suppress console output

filipesoliveira:
The output generated to the stdout and stderr by a Gramine enclave in release mode is discarded.


filipesoliveira:
The ticket https://r3-cev.atlassian.net/browse/CON-1283 was created so we do not forget to enable the Q.A. test to ensure no output is leaked when using a Gramine enclave in release mode.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -71,6 +72,13 @@ class GramineEnclaveHandle(
         return process.inputStream.reader().use { it.readText() }.trimEnd()
     }
 
+    // ProcessBuilder.Redirect.DISCARD
```
bonifazi-r3, line 75: I don't understand the comment! Where is `Redirect.DISCARD` used?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -71,6 +72,13 @@ class GramineEnclaveHandle(
...
+    // was introduce in Java 9 but Enclave still supports Java 8.
+    // That is why NULL_FILE was added in. The code was copied from ProcessBuilder.Redirect.DISCARD
+    private val NULL_FILE: File = File(
```
bonifazi-r3, line 78: Can this go in the companion object? Also, is there a missing space before "NUL"?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -71,6 +72,13 @@ class GramineEnclaveHandle(
...
+        if (System.getProperty("os.name")
+                .startsWith("Windows"))"NUL" else "/dev/null"
```
bonifazi-r3, line 80: Also, I am confused. How are we going to run Gramine in Windows? How?

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
filipesoliveira: `Redirect.DISCARD` should be used instead of `NULL_FILE` but Conclave must support Java 8 and `Redirect.DISCARD` was only introduced in Java 9.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
filipesoliveira: This block of code was copied from the Java library. It should not be changed so people know that it can be replaced by `Redirect.DISCARD`.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: Ok

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
bonifazi-r3: My opinion is that a comment is enough and we should not mention Windows here.

filipesoliveira:
The solution in this PR might not be valid after a quick discussion with Alex, and Marco.

shamsasari (changes requested):
Whilst this is a nice solution, it's not workable in an SGX context. The host could be malicious and can simply start the enclave process with the outputs turned on. The enclave itself must be turned stdout/stderr off.

Setting `System.setOut`/`setErr` may be good enough for now.

filipesoliveira:
Replaced by https://github.com/R3Conclave/conclave-core-sdk/pull/143/files

PR #143, opened 2023-01-05
CON-1228 - Release mode Gramine enclave to suppress console output

filipesoliveira:


conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -244,6 +247,24 @@ abstract class Enclave {
         this.env = env
         initCryptography()
 
+        // Prevent users from printing to the console when running the enclave in release mode
+        // GraalVM already discards the output so the following if statement is only relevant for GramineSGX
+        // N.B. The user can still override the configuration below and print to the console while running
+        // the enclave in release mode. But that has to be a conscious decision.
+        if (env.enclaveMode == EnclaveMode.RELEASE && env is GramineSGXEnclaveEnvironment) {
+            // The object should be nullOutputStream instead but Conclave must support Java 8
+            val discardOuput = PrintStream(object : OutputStream() {
```
shamsasari, line 256: `nullOut` might be a better name

shamsasari: Better to have a named class

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -244,6 +247,24 @@ abstract class Enclave {
...
+                @Throws(IOException::class)
```
shamsasari, line 257: These `@Throws` are not needed

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -244,6 +247,24 @@ abstract class Enclave {
         this.env = env
         initCryptography()
 
+        // Prevent users from printing to the console when running the enclave in release mode
+        // GraalVM already discards the output so the following if statement is only relevant for GramineSGX
+        // N.B. The user can still override the configuration below and print to the console while running
+        // the enclave in release mode. But that has to be a conscious decision.
```
shamsasari, line 253: I'm not sure the final statement is correct. How can the user enclave re-apply the stdout again? I don't think they can, but let me how they could.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -244,6 +247,24 @@ abstract class Enclave {
...
+            val discardOuput = PrintStream(object : OutputStream() {
+                @Throws(IOException::class)
+                override fun write(b: Int) {
+                }
+                @Throws(IOException::class)
+                override fun write(b: ByteArray, off: Int, len: Int) {
+                }
+            })
+            System.setOut(discardOuput);
+            System.setErr(discardOuput);
```
shamsasari, line 265: Let's set this before the `initCryptography`

shamsasari (changes requested):

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
filipesoliveira: The last sentence might not be clear. By user, I am referring to the developer of the enclave. Please let me know if that is the case and I will update the last sentence to become more explicit.

The code to redirect the output again to the stdout is `System.setOut(PrintStream(FileOutputStream(FileDescriptor.out)))`.

conclave-enclave/src/main/kotlin/com/r3/conclave/enclave/Enclave.kt
```diff
@@ -242,6 +245,18 @@ abstract class Enclave {
     @PotentialPackagePrivate
     private fun initialise(env: EnclaveEnvironment) {
         this.env = env
+
+        // Prevent users from printing to the console when running the enclave in release mode
+        // GraalVM already discards the output so the following if statement is only relevant for GramineSGX
+        // N.B. The developer of the enclave can still override the configuration below and print to the console
+        // when running the enclave in release mode. But that has to be a conscious decision.
```
shamsasari, line 252: This last statement is not correct. Can you show me how they could do that

filipesoliveira: `package com.r3.conclave.sample.enclave

import java.io.FileDescriptor
import java.io.FileOutputStream
import java.io.PrintStream

/**
 * Simply reverses the bytes that are passed in.
 */
class ReverseEnclave : com.r3.conclave.enclave.Enclave() {
    // We store the previous result to showcase that the enclave internals can be examined in a mock test.
    lateinit var previousResult: ByteArray
    protected override fun receiveFromUntrustedHost(bytes: ByteArray): ByteArray? {
        // This is used for host->enclave calls so we don't have to think about authentication.
        val input = String(bytes)
        val result: ByteArray = reverse(input).toByteArray()
        previousResult = result
        println("Hello. Stdout")
        java.lang.System.err.println("Hello. Stderr")
        return result
    }

    @kotlin.Throws(java.lang.Exception::class)
    protected override fun receiveMail(mail: com.r3.conclave.mail.EnclaveMail, routingHint: String?) {
        // This is used when the host delivers a message from the client.
        // First, decode mail body as a String.
        val stringToReverse = String(mail.getBodyAsBytes())
        // Reverse it and re-encode to UTF-8 to send back.
        val reversedEncodedString: ByteArray = reverse(stringToReverse).toByteArray()
        // Get the post office object for responding back to this mail and use it to encrypt our response.
        val responseBytes: ByteArray = postOffice(mail).encryptMail(reversedEncodedString)
        postMail(responseBytes, routingHint)
    }

    companion object {
        @kotlin.Throws(java.lang.Exception::class)
        private fun reverse(input: String): String {
            System.setOut(PrintStream(FileOutputStream(FileDescriptor.out)))
            println("Hello. Stdout")
            java.lang.System.err.println("Hello. Stderr")
            if(true) {
                throw  Exception("askjdhfksd")
            }
            val builder = StringBuilder(input.length)
            for (i in input.length - 1 downTo 0) builder.append(input[i])
            return builder.toString()
        }
    }
}`

shamsasari: The reason I say that is bc the user doesn't have a reference to the stdout/stderr objects to call the set methods with to resign stdout/stderr.

shamsasari: I see. I didn't know about `FileDescriptor.out`. Fair enough.

shamsasari: In that case, can we also turn off the enclave process redirect to the host. It's not fully secure, but it at least avoids issues if the user is using Conclave as-is and tries to do this trick.

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/gramine/GramineEnclaveHandle.kt
```diff
@@ -91,11 +91,14 @@ class GramineEnclaveHandle(
...
 
-        gramineProcess = ProcessBuilder()
-            .inheritIO()
+        val gramineProcessBuilder = ProcessBuilder()
             .directory(workingDirectory.toFile())
             .command(command)
-            .start()
+
+        if(enclaveMode != EnclaveMode.RELEASE)
+            gramineProcessBuilder.inheritIO()
```
shamsasari, line 99: space after `if` and braces please.

But more importantly, can you add a comment why this is here.

shamsasari (approved):
Thanks Filipe.

PR #144, opened 2023-01-05
[NOTICK] Fix hang at enclave startup

ac101m:
# Changes
- Fix a hang that occurs if gramine fails before connecting to the host.


shamsasari:
Integration tests failing

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -55,19 +57,42 @@ class SocketHostEnclaveInterface : CallInterface<EnclaveCallType, HostCallType>(
...
             stateManager.transitionStateFrom<State.Ready>(to = State.Running)
 
             try {
                 /** Wait for IO sockets, then close the server socket. */
                 serverSocket.use {
+                    /**
+                     * Attempt to receive the maximum number of concurrent calls from the enclave.
+                     * Check the enclave subprocess periodically to ensure that it's still alive.
+                     */
+                    var connectionSuccessful = false
```
shamsasari, line 76: Use AtomicBoolean to avoid the data visibility issue

conclave-host/src/main/kotlin/com/r3/conclave/host/internal/SocketHostEnclaveInterface.kt
```diff
@@ -55,19 +57,42 @@ class SocketHostEnclaveInterface : CallInterface<EnclaveCallType, HostCallType>(
...
 
-                    /** Receive the maximum number of concurrent calls from the enclave. */
-                    it.accept().use { initialSocket ->
-                        maxConcurrentCalls = DataInputStream(initialSocket.getInputStream()).readInt()
+                    val initialConnectionThread = Thread {
+                        it.accept().use { initialSocket ->
```
shamsasari, line 79: This `it` is quite removed from where it's defined. perhaps give it a name to make it clearer.

shamsasari (changes requested):

shamsasari (approved):
Thanks

PR #145, opened 2023-01-06
CON-1290 - Updated docs

filipesoliveira:


docs/docs/api-changes.md
```diff
@@ -11,6 +11,9 @@ The default value for the Conclave configuration field `maxThreads` has been inc
 to cover most of the use-cases when the logic inside the enclaves uses a large number of threads. There should be no
 major performance impact.
 
+The new path of `signing_material.bin`, which is generated when the enclave is being prepared to be signed, is
+`enclave/build/enclave/release`. Notice that the name of the directory `release` is no longer capitalized.
```
shamsasari, line 15: It might be good to read the section on `signingMaterial` in https://docs.conclave.net/signing.html#external-key.

Maybe something like

The default path for the `signingMaterial` when using an external signing process has changed to be all lowercase. For example, previously it used to be `build/enclave/Release/signing_material.bin`, now it is `build/enclave/release/signing_material.bin`.

shamsasari: It might also be good to add a reference to the page, perhaps on the "signingMaterial" word.

shamsasari (approved):
Good catch on the docs.

PR #150, opened 2023-01-10
CON-1285: Allow back the usage of signing key outside of enclaves

bonifazi-r3:
In previous PRs we have forced the signing keys to be defined inside the enclave directories. This was due to Gramine tests that needed such change.
In this PR we reverted back to the previous behaviour, so that we can now use signing key files which are outside the enclave paths.


plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -154,12 +155,15 @@ open class GenerateGramineBundle @Inject constructor(
      * This will create a .manifest.sgx and a .sig files into the output dir.
      */
     private fun generateSgxManifestAndSigstruct() {
+        signingKey.copyToOutputDir(SIGNING_KEY)
```
anien, line 158: Can't you reuse `prepareFile` in `LinuxExec`? It does copy file into the temp dir within the project. Not sure what location is required when creating the manifest. You also have the `cleanPreparedFiles` to delete the file.

anien (changes requested):

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: It seems to work with your suggested approach too
```
Starting process 'command 'docker''. Working directory: /home/marco/conclave-tutorials/hello-world/enclave Command: docker run --rm -u 1000:1000 -v /home/marco/conclave-tutorials/hello-world/enclave:/project -w /project/build/conclave/debug/gramine-bundle conclave-build:latest gramine-sgx-sign --manifest=java.manifest --key=/project/.linuxexec/sample_private_key14535025473192576462pem --output=java.manifest.sgx
Successfully started process 'command 'docker''
```

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
```diff
@@ -156,12 +156,15 @@ open class GenerateGramineBundle @Inject constructor(
...
+        val signingKeyPath = linuxExec.prepareFile(signingKey.get().asFile.toPath())
+
         execCommand(
             "gramine-sgx-sign",
             "--manifest=${GRAMINE_MANIFEST}",
-            "--key=${signingKey.get().asFile.absolutePath}",
+            "--key=${signingKeyPath.absolutePathString()}",
             "--output=${GRAMINE_SGX_MANIFEST}"
         )
+        linuxExec.cleanPreparedFiles()
```
anien, line 167: It would be better if we surround it with `try/finally` block as we do elsewhere, e.g. `AddEnclaveSignature.kt`

anien (changes requested):

plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/gramine/GenerateGramineBundle.kt
bonifazi-r3: done, thanks!

PR #156, opened 2023-01-16
Release 1.4-BETA

bonifazi-r3:


bonifazi-r3:
@shamsasari Do we need something like this?

shamsasari (commented):
Yes, but with two changes. lowercase beta, and with a number at the end. So "1.4-beta1"

versions.gradle
```diff
@@ -1,5 +1,5 @@
 ext {
-    conclave_version = '1.4-SNAPSHOT'
+    conclave_version = '1.4-beta1'
```
shamsasari, line 2: revert for now.

shamsasari (changes requested):

shamsasari (approved):
Thanks.

PR #157, opened 2023-01-24
Change error message for GraalVM if something goes wrong with docker, replace docs links to point to the GitHub wiki

anien:


plugin-enclave-gradle/src/main/kotlin/com/r3/conclave/plugin/enclave/gradle/LinuxExec.kt
```diff
@@ -68,11 +69,11 @@ open class LinuxExec @Inject constructor(objects: ObjectFactory, private val isP
         try {
             return commandLine(dockerCommand, commandLineConfig)
         } catch (e: Exception) {
-            val message = if (OperatingSystem.current().isLinux) {
+            val message = if (OperatingSystem.current().isLinux && runtimeType.get() == GRAALVM) {
```
shamsasari, line 72: If the runtime is gramine, then this will be false, and the else will be executed, since it not correct. I think the runtime check should be nested inside. something like

```
if (OperatingSystem.current().isLinux) {
   if (runtimeType.get() == GRAALVM) {
// Conclave requires Docker to be installed when building enclaves. ....
   } else {
     throw e;
  }
}
....
```

anien: Yep, just thought about that too.

anien: PTAL

shamsasari (approved):
Thanks Anna

